PR #10, opened 2022-07-27
local.conf.sample: restore default jobs/threads

kergoth:
This reduces the quite aggressive default from `1.5*NUM_CORES` and `2.0*NUM_CORES` to `NUM_CORES` for both variables, which is oe-core's default.


abelal (approved):
We'd still be doing https://github.com/MentorEmbedded/meta-mentor/pull/1474, right? i.e. the heuristics changes on the new rebranded repository.

kergoth:
Exactly, yes, I'll still be pulling that over in a new PR.

PR #13, opened 2022-07-28
Implement jobs/threads scaling support

kergoth:
A new class is added warns if memory-per-cpu-core is low, and set
BB_NUMBER_THREADS_SCALED and PARALLEL_MAKE_SCALED with scaled down
values.

Class behavior is configured using the MEM_PER_CPU_SCALING_THRESHOLDS
variable. By default, if we have less than 4gb per CPU core, we warn and
reduce the make jobs to cpus/2. If we have less than 2gb per CPU core,
we warn and reduce both bitbake threads and make jobs to cpus/2.

An example of how to set the variables to the scaled values is provided in
local.conf.sample.

JIRA: SB-20030

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -92,6 +92,10 @@ USER_FEATURES += "~nfc"
 # How many jobs 'make' should run in parallel when compiling. Default: NUM_CPUS
 #PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count()}"
 
+# Uncomment to scale down threads and jobs when memory-per-cpu-core is low
+#BB_NUMBER_THREADS = "${BB_NUMBER_THREADS_SCALED}"
+#PARALLEL_MAKE = "${PARALLEL_MAKE_SCALED}"
+
 # Uncomment to exclude GPLv3 software from the build
 #INCOMPATIBLE_LICENSE = "GPL-3.0* LGPL-3.0*"
 
--- a/meta-sokol-flex-distro/classes/mem-per-cpu-scaling.bbclass
+++ b/meta-sokol-flex-distro/classes/mem-per-cpu-scaling.bbclass
@@ -1,0 +1,137 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+#
+# Warn if memory-per-cpu-core is low, and set BB_NUMBER_THREADS_SCALED and
+# PARALLEL_MAKE_SCALED with scaled down values.
+#
+# Inspired by https://elinux.org/images/d/d4/Goulart.pdf.
+#
+# Configured using the MEM_PER_CPU_SCALING_THRESHOLDS variable:
+#
+# MEM_PER_CPU_THRESHOLD,THREADS,JOBS,ACTION
+#
+# If mem per cpu is less than MEM_PER_CPU_THRESHOLD, set BB_NUMBER_THREADS_SCALED to
+# THREADS, PARALLEL_MAKE_SCALED to JOBS, and execute specified ACTION if set. If THREADS
+# or JOBS are empty, they are left unmodified.
+MEM_PER_CPU_SCALING_THRESHOLDS ?= "\
+    4,cpus,cpus//2,WARN \
+    2,cpus//2,cpus//2,WARN \
+"
+
+BB_NUMBER_THREADS_SCALED = "1"
+PARALLEL_MAKE_SCALED = "-j 1"
+
+# This function postpones warnings to reduce duplication, and postpones failures
+# to avoid breaking bitbake -e.
+python mem_per_cpu_scaling() {
+    """Set bitbake threads and make jobs based upon available memory per CPU core."""
+
+    current_threads = d.getVar('BB_NUMBER_THREADS', False)
+    threads_are_scaled = current_threads == '${BB_NUMBER_THREADS_SCALED}'
+    if not threads_are_scaled:
+        current_threads = int(d.expand(current_threads))
+
+    current_jobs = d.getVar('PARALLEL_MAKE', False)
+    jobs_are_scaled = current_jobs == '${PARALLEL_MAKE_SCALED}'
+    if not jobs_are_scaled:
+        current_jobs = int(d.expand(current_jobs).split()[1])
+
+    try:
+        matched, mem_per_cpu, warning = matching_rule(d)
+    except ValueError as exc:
+        d.setVar('MEM_PER_CPU_SCALING_ERROR', str(exc))
+        return
+    else:
+        if warning:
+            d.setVar('MEM_PER_CPU_SCALING_WARNING', warning)
+
+        if not matched:
+            return
+        else:
+            threshold, threads, jobs, action = matched
+    
+    suggested = []
+    if threads and (threads_are_scaled or current_threads > threads):
+        if not threads_are_scaled:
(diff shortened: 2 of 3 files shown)


kergoth:
@abelal When migrating this, I realigned the config based on your request, which is less aggressive, and corrected the typo you noticed.

kergoth:
Updated to add example output messages to the commit message.

kergoth:
Updated the commit message further to include a mention of the variable that controls the behavior, as well as the default configuration used.

ahsanhussain:
LGTM

kergoth:
[**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Aug 3, 2022, 4:24 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1203819251 "2022-08-03T11:24:26Z - Replied by Github Reply Comments"):
> LGTM

Hmm, I was looking at this and realizing that it also sets the jobs/threads default value, not only in the case where we need to scale down. I'm wondering if I should change this to only alter the already set values. That is, the user can currently override the jobs and threads vars in local.conf, but their default values aren't currently coming from oe-core, they're also set by the variable that controls this.

Thoughts? Should it only change jobs/threads when we're scaling down, or also set the default value?

abelal:
> [**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Aug 3, 2022, 4:24 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1203819251):
> 
> > LGTM
> 
> Hmm, I was looking at this and realizing that it also sets the jobs/threads default value, not only in the case where we need to scale down. I'm wondering if I should change this to only alter the already set values. That is, the user can currently override the jobs and threads vars in local.conf, but their default values aren't currently coming from oe-core, they're also set by the variable that controls this.
> 
> Thoughts? Should it only change jobs/threads when we're scaling down, or also set the default value?

@kergoth I think we should pick the defaults from oe-core and only scale down when there's a need.

kergoth:
[**@abelal**](https://github.com/abelal) commented on [Aug 3, 2022, 8:48 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1204136014 "2022-08-03T15:48:58Z - Replied by Github Reply Comments"):
> > [**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Aug 3, 2022, 4:24 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1203819251):
> > 
> > > LGTM
> > 
> > Hmm, I was looking at this and realizing that it also sets the jobs/threads default value, not only in the case where we need to scale down. I'm wondering if I should change this to only alter the already set values. That is, the user can currently override the jobs and threads vars in local.conf, but their default values aren't currently coming from oe-core, they're also set by the variable that controls this.
> > 
> > Thoughts? Should it only change jobs/threads when we're scaling down, or also set the default value?
> 
> @kergoth I think we should pick the defaults from oe-core and only scale down when there's a need.

That was my thought as well, I'll test that out. Thanks

kergoth:
@abelal Please re-review.

abelal (approved):

Noor-Ahsan:
Missing license header from bbclass. See example at https://github.com/MentorEmbedded/meta-mentor/blob/dunfell/meta-mel-support/classes/image-sanity-incompatible.bbclass#L1

kergoth:
[**@kergoth**](https://github.com/kergoth) commented on [Aug 4, 2022, 7:56 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1205371839 "2022-08-04T14:56:10Z - Replied by Github Reply Comments"):
> > Missing license header from bbclass. See example at [https://github.com/MentorEmbedded/meta-mentor/blob/dunfell/meta-mel-support/classes/image-sanity-incompatible.bbclass#L1](https://github.com/MentorEmbedded/meta-mentor/blob/dunfell/meta-mel-support/classes/image-sanity-incompatible.bbclass?rgh-link-date=2022-08-04T12%3A24%3A53Z#L1)
> 
> Good eye, thanks!

Added license header. Thanks.

Noor-Ahsan:
@kergoth Just for my understanding this will update the values that we have set in local.conf? for example if we have folowing in local.conf the implementation above will overwrite it?

BB_NUMBER_THREADS ?= "${@int(oe.utils.cpu_count() * 1.5)}"
PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count() * 2}"

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Aug 12, 2022, 12:01 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1212793145 "2022-08-12T07:01:39Z - Replied by Github Reply Comments"):
> [@kergoth](https://github.com/kergoth) Just for my understanding this will update the values that we have set in local.conf? for example if we have folowing in local.conf the implementation above will overwrite it?
> 
> BB\_NUMBER\_THREADS ?= "${[@int](https://github.com/int)(oe.utils.cpu\_count() \* 1.5)}"  
> PARALLEL\_MAKE ?= "-j ${@oe.utils.cpu\_count() \* 2}"

The original implementation would not, the new one will, but only in the case where resources are limited. Maybe we should think about the semantics again. The intention of having it not define it for the default case was to allow the oe-core default to take effect when we aren't scaling down, but as a consequence of the implementation change, now it overrides user configuration when scaling.

abelal:
> [**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Aug 12, 2022, 12:01 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1212793145):
> 
> > [@kergoth](https://github.com/kergoth) Just for my understanding this will update the values that we have set in local.conf? for example if we have folowing in local.conf the implementation above will overwrite it?
> > BB_NUMBER_THREADS ?= "${[@int](https://github.com/int)(oe.utils.cpu_count() * 1.5)}"
> > PARALLEL_MAKE ?= "-j ${@oe.utils.cpu_count() * 2}"
> 
> The original implementation would not, the new one will, but only in the case where resources are limited. Maybe we should think about the semantics again. The intention of having it not define it for the default case was to allow the oe-core default to take effect when we aren't scaling down, but as a consequence of the implementation change, now it overrides user configuration when scaling.

Hmmm. The MR says "help the customer", maybe we could just warn in case of resource limitations and provide an appropriate suggestion rather than overriding? We could also provide a variable through local.conf which could manipulate this behavior, sounds like an overkill / over-worked feature?

kergoth:
@abelal @Noor-Ahsan Reworked to simplify and set optionally used variables. I also changed the default scaling values, as expecting 8gb per core seems unrealistic.

abelal:
If the build is using scaled values we shouldn't generate the warning.

kergoth:
[**@abelal**](https://github.com/abelal) commented on [Sep 6, 2022, 8:40 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/13#issuecomment-1238323816 "2022-09-06T15:40:10Z - Replied by Github Reply Comments"):
> If the build is using scaled values we shouldn't generate the warning.

Updated. Now we only show messages if we would be changing the values, and we never show anything if the user has already switched to use the scaled values, or if they've manually corrected them. It also shows a more detailed message with the suggested change:
```
WARNING: Low memory per CPU core (~3 Gb) in this system. You may wish to reduce PARALLEL_MAKE to 16 or less in local.conf.
```

abelal (approved):

PR #16, opened 2022-07-29
udev: fix for systemd mounting timeout issue for dmsetup encrypted pa…

paw3142:
…rtition

Source: https://bugs.gentoo.org/706434

Signed-off-by: Amy Fong <Amy_Fong@mentor.com>

### DIFF
--- a/meta-sokol-flex-distro/openembedded-layer/recipes-support/lvm2/lvm2/09-dm_hack.rules
+++ b/meta-sokol-flex-distro/openembedded-layer/recipes-support/lvm2/lvm2/09-dm_hack.rules
@@ -1,0 +1,8 @@
+KERNEL=="device-mapper", NAME="mapper/control"
+SUBSYSTEM!="block", GOTO="dm_end"
+KERNEL!="dm-[0-9]*", GOTO="dm_end"
+ACTION!="add|change", GOTO="dm_end"
+
+ENV{DM_UDEV_PRIMARY_SOURCE_FLAG}="1"
+
+LABEL="dm_end"
--- a/meta-sokol-flex-distro/openembedded-layer/recipes-support/lvm2/lvm2_%.bbappend
+++ b/meta-sokol-flex-distro/openembedded-layer/recipes-support/lvm2/lvm2_%.bbappend
@@ -5,6 +5,7 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 SRC_URI:append = " file://lvm2-flex.service"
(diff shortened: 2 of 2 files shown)


paw3142:
Please review @chombourger @Noor-Ahsan @abelal

kergoth:
Can you add a note about what this rules file changes relative to the default behavior that fixes the timeout to the commit message? It's not entirely clear which change works around the problem when reading the rules file, at least to my eyes. Thanks!

paw3142:
It has to do with DM_UDEV_PRIMARY_SOURCE_FLAG (set in 10-dm.rules)

I just stumbled upon this which explains the situation:

https://lore.kernel.org/all/CANnVG6mMTcRkWQEEhZSEELfyGHqNkGPx2LmdR9NS+-wkAieGSw@mail.gmail.com/T/

kergoth:
[**@paw3142**](https://github.com/paw3142) commented on [Aug 1, 2022, 7:04 PM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/16#issuecomment-1201931355 "2022-08-02T02:04:56Z - Replied by Github Reply Comments"):
> It has to do with DM_UDEV_PRIMARY_SOURCE_FLAG (set in 10-dm.rules)
> 
> I just stumbled upon this which explains the situation:
> 
> https://lore.kernel.org/all/CANnVG6mMTcRkWQEEhZSEELfyGHqNkGPx2LmdR9NS+-wkAieGSw@mail.gmail.com/T/

Thanks, would you mind adding a short note about that to the commit message (see commit --amend)?

kergoth (changes requested):
The content looks alright to me, but if this is a fix, not a distro specific workaround, should it go upstream? Does it belong more in meta-sokol-flex-staging rather than meta-sokol-flex-distro? Also, this is targeting the wrong branch if you want this to end up in the Sokol Flex OS 13 release, we're using kirkstone, not main.

paw3142:
I'd be hesitant to send it upstream since it is a hack to make things work for customized initrds -- from the thread in lore.kernel.org, there's some userspace patches for systemd/udevd (https://github.com/systemd/systemd/pull/13881) that seems to be a better path of fixing the problem.

I'm not sure about the correct location for this fix (mel-flex newbie here). I created the fix in a customized branch and I'm trying to send it up the chain.... 

Let me know if I should sent it somewhere else. Thanks!

abelal:
Replaced by > https://github.com/MentorEmbedded/meta-sokol-flex/pull/126

PR #17, opened 2022-07-29
Use TOOLCHAINS_PATH in local.conf.sample directly

kergoth:
This aligns with how we do things, is simpler, in that it avoids changing variable names, and doesn't imply the user should use the wrong variable if they read `local.conf.sample` directly.


abelal (approved):

kergoth:
https://github.com/MentorEmbedded/meta-sokol-flex/pull/20 supercedes most of this. By removing the logic from the setup scripts, there's no need to correct that logic.

PR #19, opened 2022-08-02
sokol-flex: default to internal toolchain until we have a usable toolchain delivery

kergoth:


meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -51,9 +51,6 @@ WARN_TO_ERROR_QA ??= ""
 WARN_QA:remove = "${WARN_TO_ERROR_QA}"
 ERROR_QA:append = " ${WARN_TO_ERROR_QA}"
 
-# Default to the external toolchain if available
-TCMODE = "${@bb.utils.contains('BBFILE_COLLECTIONS', 'sourcery', 'external-sourcery-rebuild-libc', bb.utils.contains('BBFILE_COLLECTIONS', 'external-toolchain', 'external', 'default', d), d)}"
```
abelal, line 55: Rather than including meta-sourcery in the repo setup and removing this I believe a simpler way would be to not include meta-sourcery in the metadata for now which would automatically switch to the internal tools. @kergoth, thoughts?

Once you are done with the integration and updates from your local setup we can simply pick up meta-sourcery and that would enable the external tools automatically.

PR #20, opened 2022-08-02
Set TOOLCHAINS_PATH via the metadata

kergoth:
There's no need to hardcode this knowledge in the setup scripts when we can set it in the metadata instead.

This makes https://github.com/MentorEmbedded/meta-sokol-flex/pull/17 unnecessary.

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -6,9 +6,9 @@ DISTRO = 'sokol-flex'
 # qemuarm, qemumips, qemuppc, qemux86, qemux86-64.
 MACHINE ??= "qemux86"
 
-# The path to the installed sourcery toolchain. This defaults to using the one
-# from the CodeBench we're installed in.
-#EXTERNAL_TOOLCHAIN ?= "/path/to/toolchain"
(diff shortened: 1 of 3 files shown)


conf/local.conf.sample
```diff
@@ -6,9 +6,8 @@ DISTRO = 'sokol-flex'
 # qemuarm, qemumips, qemuppc, qemux86, qemux86-64.
 MACHINE ??= "qemux86"
 
-# The path to the installed sourcery toolchain. This defaults to using the one
-# from the CodeBench we're installed in.
-#EXTERNAL_TOOLCHAIN ?= "/path/to/toolchain"
+# The path to the toolchains installed with CodeBench.
+#TOOLCHAINS_PATH ?= "/path/to/toolchains"
```
abelal, line 10: Would it make sense to clarify what the default would be in case it isn't set here?

abelal:
This definitely is a better approach :+1:

conf/local.conf.sample
kergoth: Yes, good idea, thanks. I'll update it.

kergoth: Updated, thanks!

abelal (approved):
:+1:

PR #21, opened 2022-08-03
Switch to BBFILES_DYNAMIC & dynamic-layers

kergoth:
This aligns with upstream layer conventions, but generates it rather than
explicitly listing each, for easier maintenance.

JIRA: SB-20093


kergoth:
After we review this, we can apply the same pattern to the other layers.

abelal (approved):
Cool!

PR #27, opened 2022-08-10
Prune the unnecessary from our .sample template files

kergoth:
# Prune the unnecessary from our .sample template files

Our intention here is to minimize the `local.conf.sample` to the set of customizations in our product documentation along with a few necessary exceptions and the appends from the layers. Any examples of customization beyond our documentation can be found in the referenced oe-core local.conf.sample and local.conf.sample.extended, or the Yocto Project documentation. This should minimize our local.conf and make it less confusing for a new user.

## Keep in local.conf despite not being in the docs

- `SOURCERY_VERSION_REQUIRED`: This ensures that the used external toolchain version aligns with the expectations for this product release. It's a good sanity check to have, and we want the user to have visibility into it, so they can override it if necessary, but at the same time the user isn't expected to customize it under normal circumstances, so it should not be documented. **Update** this is no longer in the `local.conf.sample` in the layer, but is appended to the template local.conf.sample which is shipped with the release, the same way we handle `PDK_DISTRO_VERSION`.

## Candidates for moving outside of local.conf

- [x] BB_DISKMON_DIRS: move to sokol-flex.conf
- [x] `drop-toolchain-from-sdk.inc`: move to `packagegroup-cross-canadian.bbappend` for our distro, to be revisited when we gain external toolchain redistribution

## Candidates for removal

- [x] IMAGE_FSTYPES: the BSP handles this, and those are converging on wic anyway, so a customer should not need this
- [x] Removal of 3g and nfs: this is arbitrary, we shouldn't impose preconceptions of particular usages of our product
- [x] TMPDIR: more common is to alter DL_DIR and SSTATE_DIR to site-specific locations. the one case I can see for this is where the user directs all build output to a writable path if their workspace is not writable, but that can be dealt with by constructing the workspace in a writable location.
- [x] PATCHRESOLVE: this mechanism is both rarely used and not well supported.
- [x] SDK_TITLE, SDK_IDENTIFIER: these are much less visible than previous and are unlikely to be customized, in my opinion
- [x] copyleft_compliance: this is old and much less capable than archiver, and archive-release doesn't use it anymore, either, in favor of its own archive-release-downloads class. It's also not aligned with upstream
- [x] SDK_PACKAGING_FUNC: our docs will now be covering use of SDK installers, not bare tarballs except in the case of Windows SDKMACHINE, so I don't think we need an example of changing this
- [x] `PACKAGE_CLASSES`
- [x] `OE_TERMINAL`
- [x] `IMAGE_GEN_DEBUGFS` and `EXTRA_DEBUG_PACKAGES`
- [x] Install tzdata with systemd when building read only images.  This workaround is no longer required.
- [x] `PREMIRRORS`: The uncommented usages which currently exist are done to support a read-only `FLEXDIR`, which likely isn't something we have to support given git-based delivery and the creation of Sokol Flex workspaces. That is, the customer can create a workspace in a writable location if their Sokol Flex installation is read-only, rather than supporting a read-only workspace with a writable build directory.
- [x] gdbserver-gplv3 and binutils license exceptions, in favor of `INCOMPATIBLE_LICENSE_EXCEPTIONS` example

## Candidates for additions to our documentation

- ptest, for runtime testing
- archiver, for license compliance
- `SDKMACHINE`: This should be in the forthcoming Yocto SDK documentation section, to allow for Windows-native SDK builds
- I think we should document how to add packages to the SDK in our new Yocto SDK documentation section. This should cover our change to the default `TOOLCHAIN_HOST_TASK` as well as how to add packages to the Yocto SDK via the available variables to do so. I believe it's common enough to want to customize SDK contents that it's worth documenting.
- `own-mirrors`: This is used quite often, and we even use it for our own developers to pull from an internal mirror. It's also analogous to the already-documented `SSTATE_MIRROR_SITES`.
- `INCOMPATIBLE_LICENSE_EXCEPTIONS` should be covered where we cover `INCOMPATIBLE_LICENSE`.

<details>
<summary>References in our documentation</summary>

### "Building the target image"

- `ROOT_PASSWORD`
- `MULTILIBS`

### "Building With the Internal Toolchain"

- `TCMODE`

### "Build Customization"

- `DISTRO`
- `MACHINE`
- `EXTRA_IMAGE_FEATURES`
- `CORE_IMAGE_EXTRA_INSTALL`
- `BB_NO_NETWORK`
- `BB_NUMBER_THREADS`
- `PARALLEL_MAKE`
- `INCOMPATIBLE_LICENSE`
- `DL_DIR`
- `SSTATE_DIR`
- `USER_CLASSES`
- `SSTATE_MIRROR_SITES`
- `IMAGE_FEATURES_DEVELOPMENT`
- `IMAGE_FEATURES_PRODUCTION`

#### Secure Boot

- `SECURE_BOOT_PKI_DIR`

### eMMC Boot Workflow

- `SERVICE_IMAGE_BUNDLE_IMAGES`
- `BOOT_MEDIA`
- `ACCEPT_FSL_EULA`

#### collectd

- `COLLECTD_INFLUXDB_SERVER_ADDR`
- `COLLECTD_INFLUXDB_SERVER_PORT`

### software update

- `SWUPDATE_SIGN_PUBLIC_KEY`
- `SWUPDATE_VERBOSE`
- `SWU_KEYS_PATH`
- `SWUPDATE_ENC_INFO`
- `HAWKBIT_SERVER_URL`
- `HAWKBIT_GW_TOKEN`
- `SWUPDATE_BOOT_LIMIT`
- `DISABLE_SWUPDATE_SERVICE`
- `SWU_IMAGE_VERSION`
</details>


kergoth:
I need input on this, especially on the Uncertain and Candidates for doc additions.

kergoth:
Updated to exclude toolchain by altering `packagegroup-cross-canadian` rather than brute forcing its removal, as it's a royal pain to alter the default `TOOLCHAIN_HOST_TASK` or `TOOLCHAIN_TARGET_TASK` due to recipes and `populate_sdk_base` both defining them with `?=`.

conf/conf-notes.txt
```diff
@@ -1,6 +1,5 @@
 Common targets are:
 
-    core-image-base
```
abelal, line 3: hmmmm not sure about this. @Noor-Ahsan ?

conf/local.conf.sample
```diff
@@ -19,8 +17,8 @@ MACHINE ??= "qemux86"
 # development-image and production-image..
 #CORE_IMAGE_EXTRA_INSTALL += "bash"
 
-# The EXTRA_IMAGE_FEATURES variable allows groups of packages to be added to
-# the generated images. Some of these options are added to certain image types
+# The EXTRA_IMAGE_FEATURES variable allows groups of packages to be added all
+# generated images. Some of these options are added to certain image types
```
abelal, line 25: to be added _to_ all generated images.

conf/local.conf.sample
```diff
@@ -74,7 +74,7 @@ CORE_IMAGE_EXTRA_INSTALL += "${@bb.utils.contains('IMAGE_FEATURES', 'read-only-r
 #ROOT_PASSWORD = "0"
 IMAGE_CLASSES:append = " image-root-password"
 
-# User features lets you manipulate the distro features. To add a distro
+# User features lets more easily manipulate the distro features. To add a distro
```
abelal, line 77: User feature allows easier manipulation of the distro features

"lets more easily" sounds weird IMO.

conf/local.conf.sample
```diff
@@ -168,15 +168,6 @@ TOOLCHAIN_HOST_TASK:remove = "${TOOLCHAIN_HOST_REMOVE}"
...
 #TOOLCHAIN_TARGET_TASK_EXTRA += " ncurses-libncurses"
 
-# Uncomment to set a title for the SDK
-# Default: SDK for ${SDK_IDENTIFIER}
-#SDK_TITLE ?= "My Company's SDK for ${SDK_IDENTIFIER}"
-
-# Uncomment to alter the identifier for the SDK. This mechanism is used to
-# isolate the SDK on disk, as well as the installers.
-# Default: ${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${SDK_VERSION}
-#SDK_IDENTIFIER ?= "${SDKMACHINE}-${IMAGE_BASENAME}-${MACHINE}-${SDK_VERSION}.customized"
```
abelal, line 193: Why'd we have these in the first place?

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -165,6 +151,39 @@ TOOLCHAIN_TARGET_TASK:append = " ${TOOLCHAIN_TARGET_TASK_EXTRA}"
...
+BB_GIT_SHALLOW ??= "1"
+BB_GIT_SHALLOW_DEPTH_DEFAULT ??= "1"
+BB_GIT_SHALLOW_DEPTH ??= "${@'${BB_GIT_SHALLOW_DEPTH_DEFAULT}' if not '${BB_GIT_SHALLOW_REVS}' else '0'}"
+BB_GIT_SHALLOW_REVS ??= ""
+
+# For our kernel repository, we want to ship the user the history which they
+# can't easily re-acquire from upstream.
+LINUX_VERSION_TAG ??= "v${@'.'.join('${LINUX_VERSION}'.split('.')[:2])}"
+LINUX_VERSION ??= ""
+BB_GIT_SHALLOW_REVS:pn-linux-mel ??= "${@'${LINUX_VERSION_TAG}' if '${LINUX_VERSION_TAG}' and '${LINUX_VERSION}' else ''}"
```
abelal, line 185: Shouldn't this be done through the BSP layer?

meta-sokol-flex-distro/recipes-core/packagegroups/packagegroup-cross-canadian.bbappend
```diff
@@ -0,0 +1 @@
+RDEPENDS:${PN}:tcmode-external:sokol-flex = "meta-environment-${MACHINE}"
```
abelal, line 1: How do we ship the toolchain in SDKs then? I guess I am not following this correctly.

kergoth: That's not implemented yet for external toolchains. We'd drop this when we implement external toolchain redistribution.

conf/local.conf.sample
kergoth: Yes, I have no idea how this grammar got mucked up, thanks.

conf/local.conf.sample
kergoth: Originally the ADEs weren't installed into isolated locations, the identifier controlled it, so we allowed for customization to control whether it'd replace/upgrade an existing ADE or be isolated. That's since changed. And of course the ade title/sitename and identifier were both user visible in the install/remove software dialog. Now the identifier is used by codebench and is in the install path, but that's about it, and the title/sitename is only visible in the shell archive installer, and doesn't mention company name, so the user is unlikely to want to change something so trivial to their own.

meta-sokol-flex-distro/recipes-core/packagegroups/packagegroup-cross-canadian.bbappend
kergoth: This was needed to prevent building and including an internal toolchain with a built Yocto SDK/ADE while building using an external toolchain. That is, we expected a user using the external codebench toolchain to build Flex would also want an app developer to use a codebench toolchain, not a Yocto-built internal one.

kergoth:
Any thoughts on the Uncertain and Candidates items?

Noor-Ahsan:
From uncertain section I think following should be removed
OE_TERMINAL
IMAGE_GEN_DEBUGFS and EXTRA_DEBUG_PACKAGES
PACKAGE_CLASSES

We use own-mirror so we can keep it.
Where we will move PREMIRRORS if we remove it from here

Kindly let us know which other then above you need feedback.

kergoth:
> From uncertain section I think following should be removed OE_TERMINAL IMAGE_GEN_DEBUGFS and EXTRA_DEBUG_PACKAGES PACKAGE_CLASSES
> 
> We use own-mirror so we can keep it. Where we will move PREMIRRORS if we remove it from here
> 
> Kindly let us know which other then above you need feedback.

`PREMIRRORS` is an open question, from Slack:

    Hmm, we arranged for flex to write sstate and downloads to `TOPDIR` (aka the bulid dir) and still pull from `FLEXDIR` (flex install/workspace) to allow for cases where the flex layers, etc are not writable or are common. Now that we construct workspaces from the flex install, the user could easily just create the workspace in a writable location, so I'm not sure if we still need to support the case where `TOPDIR` is writable but `FLEXDIR` is not.

    I'm also not sure we need to keep the ability to pull downloads from `${LAYERDIR}/downloads`, or from `${FLEXDIR}/${MACHINE}/downloads`, since the git based delivery already keeps those artifacts separated by checksum and links them into place, so there's no conflict for p2 delivery for the installer already.

I'd also like input on the items I indicated may be worth adding to the docs. Do you agree, or think those should be dropped? I think they're common enough requirements, such as the use of archiver for license compliance, that they should be documented, but that's just my opinion.

kergoth:
@Noor-Ahsan @chombourger We have a special case to work around an aarch64 build failure in the gplv2 binutiuls when building non-gplv3. Given we aren't giving much development time or effort to the non-gplv3 builds, and you can work around it instead by excluding tools-profile from the build, I'm inclined to say we just don't support profiling in a non-gplv3 environment. Thoughts? The versions of the tools in meta-gplv2 are already ancient and falling farther behind day by day, which is why it's no longer supported in future Yocto releases, so I think the more we can minimize our support of the gplv2 build, the better.

chombourger:
> @Noor-Ahsan @chombourger We have a special case to work around an aarch64 build failure in the gplv2 binutils when building non-gplv3 (via specifically allowing the gplv3 binutils into the build for the sole purpose of profiling, much the way we do for the gplv3 gdbserver for development). Given we aren't giving much development time or effort to the non-gplv3 builds, and you can work around it instead by excluding tools-profile from the build, I'm inclined to say we just don't support profiling in a non-gplv3 environment. Thoughts? The versions of the tools in meta-gplv2 are already ancient and falling farther behind day by day, which is why it's no longer supported in future Yocto releases, so I think the more we can minimize our support of the gplv2 build, the better.

I agree. While I haven't followed latest developments in the meta-GPLv2 world, I fear that non-GPLv3 are getting every single day closer to utopia. Those ancient tools lack support for RiscV, DWARF5, etc. It's for that reason that I had asked Kathy to drop support for non-GPLv3 builds in the product. We can still entertain that discussion for SOW work where we can hopefully contain the various configurations to support.

kergoth:
> > @Noor-Ahsan @chombourger We have a special case to work around an aarch64 build failure in the gplv2 binutils when building non-gplv3 (via specifically allowing the gplv3 binutils into the build for the sole purpose of profiling, much the way we do for the gplv3 gdbserver for development). Given we aren't giving much development time or effort to the non-gplv3 builds, and you can work around it instead by excluding tools-profile from the build, I'm inclined to say we just don't support profiling in a non-gplv3 environment. Thoughts? The versions of the tools in meta-gplv2 are already ancient and falling farther behind day by day, which is why it's no longer supported in future Yocto releases, so I think the more we can minimize our support of the gplv2 build, the better.
> 
> I agree. While I haven't followed latest developments in the meta-GPLv2 world, I fear that non-GPLv3 are getting every single day closer to utopia. Those ancient tools lack support for RiscV, DWARF5, etc. It's for that reason that I had asked Kathy to drop support for non-GPLv3 builds in the product. We can still entertain that discussion for SOW work where we can hopefully contain the various configurations to support.

Thanks for the input!

I'd love to see support for it dropped, at least in its GPL-2.0 incarnation, but I don't know that management would agree in the Flex 13 timeframe. Folks are moving toward BSD/MIT-based options for non-GPL-3.0 support, which is a much more sustainable option.

kergoth:
Updated to remove the gdbserver-gplv3 and binutils license exceptions workarounds in favor of a commented example of how the customer can do so.

kergoth:
Thoughts on changing the order of items in local.conf.sample based on the order of the build customizations chapters in our documentation?

kergoth:
I will also need input on the `Candidates for additions to our documentation` items. @Noor-Ahsan @ahsanhussain @abelal @chombourger . Thanks! I believe the changes needed to local.conf are now complete with the possible exception of reordering.

kergoth:
[**@chombourger**](https://github.com/chombourger) commented on [Aug 18, 2022, 11:49 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/27#issuecomment-1219823145 "2022-08-18T18:49:06Z - Replied by Github Reply Comments"):
> > [@Noor-Ahsan](https://github.com/Noor-Ahsan) [@chombourger](https://github.com/chombourger) We have a special case to work around an aarch64 build failure in the gplv2 binutils when building non-gplv3 (via specifically allowing the gplv3 binutils into the build for the sole purpose of profiling, much the way we do for the gplv3 gdbserver for development). Given we aren't giving much development time or effort to the non-gplv3 builds, and you can work around it instead by excluding tools-profile from the build, I'm inclined to say we just don't support profiling in a non-gplv3 environment. Thoughts? The versions of the tools in meta-gplv2 are already ancient and falling farther behind day by day, which is why it's no longer supported in future Yocto releases, so I think the more we can minimize our support of the gplv2 build, the better.
> 
> I agree. While I haven't followed latest developments in the meta-GPLv2 world, I fear that non-GPLv3 are getting every single day closer to utopia. Those ancient tools lack support for RiscV, DWARF5, etc. It's for that reason that I had asked Kathy to drop support for non-GPLv3 builds in the product. We can still entertain that discussion for SOW work where we can hopefully contain the various configurations to support.

I've replaced the sections about allowing the gplv3 gdbserver and binutils for aarch64 with this:
```bitbake
# Exceptions can be made for GPL-3.0 software that's required for development or
# profiling, for example, gdbserver for debugging and binutils for lttng profiling.
#INCOMPATIBLE_LICENSE_EXCEPTIONS += "binutils:GPL-3.0-only gdbserver-external:GPL-3.0"
```

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: I've opened PRs on this repo and the meta-flex-ref-bsps repo to get this moved.

abelal (approved):

abelal:
> Thoughts on changing the order of items in local.conf.sample based on the order of the build customizations chapters in our documentation?

Lets not do that because this might then become a regular exercise :) Also ideally it would really depend on the customer use-case when they refer to docs.

abelal:
> Candidates for additions to our documentation
> 
> ptest, for runtime testing

What exactly do we want to do here? explain ptest or the inclusion? Should we simply point to the Yocto doc? https://wiki.yoctoproject.org/wiki/Ptest

> archiver, for license compliance

Makes sense IMO.

> SDKMACHINE: This should be in the forthcoming Yocto SDK documentation section, to allow for Windows-native SDK builds
> I think we should document how to add packages to the SDK in our new Yocto SDK documentation section. This should cover our change to the default TOOLCHAIN_HOST_TASK as well as how to add packages to the Yocto SDK via the available variables to do so. I believe it's common enough to want to customize SDK contents that it's worth documenting.

Again, agreed.

> own-mirrors: This is used quite often, and we even use it for our own developers to pull from an internal mirror. It's also analogous to the already-documented SSTATE_MIRROR_SITES.


I think pointing to Yocto docs would make most sense.

> INCOMPATIBLE_LICENSE_EXCEPTIONS should be covered where we cover INCOMPATIBLE_LICENSE.

+1

abelal:
Also I'd like to point out that our local.conf seems to have way too much explained essentially feeling like a user facing guide when  we should rather only have a line or two at max for each variable and discuss it thoroughly in our docs if need be.

kergoth:
[**@abelal**](https://github.com/abelal) commented on [Aug 19, 2022, 5:25 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/27#issuecomment-1220614923 "2022-08-19T12:25:56Z - Replied by Github Reply Comments"):
> Also I'd like to point out that our local.conf seems to have way too much explained essentially feeling like a user facing guide when we should rather only have a line or two at max for each variable and discuss it thoroughly in our docs if need be.

That's a good point, especially archiver is very verbose and could be trimmed down to how to enable it, and where to learn more about how to customize it. I'll look into that.

Noor-Ahsan:
@kergoth this needs rebase.

kergoth:
Rebased and updated to also pare down the comments, please review that addition.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -382,9 +382,6 @@ FLEX_EXTERNAL_ARTIFACTS ?= "${TOPDIR}/flex-external-artifacts"
 # Support for SOURCERY_VERSION_REQUIRED
 require conf/distro/include/sourcery-version-check.inc
 
-# Handle GPLv3 gdbserver
```
Noor-Ahsan, line 366: I think we need ore information in commit summary for these changes

Noor-Ahsan (commented):
I think we need more information in commit summary for this changes

Ignore this comment. I was trying to add it in a commit

meta-sokol-flex-distro/conf/distro/include/gdbserver-gplv3.inc
```diff
@@ -1,17 +0,0 @@
-# ---------------------------------------------------------------------------------------------------------------------
```
Noor-Ahsan, line 1: I think we need ore information in commit summary for these changes

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: This is already covered in the commit `local.conf.sample: remove binutils and gplv3 hacks`. This is just removing the bits the local.conf was referencing.

meta-sokol-flex-distro/conf/distro/include/gdbserver-gplv3.inc
kergoth: This is already covered in the commit `local.conf.sample: remove binutils and gplv3 hacks`. This is just removing the bits the local.conf was referencing.

kergoth: I also added a mention of this in the commit.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: I also added a mention of this in the commit.

conf/local.conf.sample
```diff
@@ -48,275 +46,83 @@ MACHINE ??= "qemux86"
...
 # Valid values: i686, x86_64, i686-mingw32, x86_64-mingw32
 #SDKMACHINE ?= "x86_64"
 
-# Uncomment to make populate_sdk write a tar file rather than a .sh installer
-#SDK_PACKAGING_FUNC = ""
-
-# By default, only include the target sysroot in the SDK, not host tools.
-# Comment this line to change that. This packagegroup includes host tools like
-# autoconf, automake, etc.
+# By default, include a toolchain the SDK, but not host tools like autoconf
```
abelal, line 108: _in_ the SDK

conf/local.conf.sample
```diff
@@ -48,275 +46,83 @@ MACHINE ??= "qemux86"
...
 TOOLCHAIN_HOST_REMOVE ??= "nativesdk-packagegroup-sdk-host"
 TOOLCHAIN_HOST_TASK:remove = "${TOOLCHAIN_HOST_REMOVE}"
 
-# TOOLCHAIN_HOST_TASK is used to add host packages to the SDK, for
-# example, to add bash:
+# Add host packages to the SDK
 #TOOLCHAIN_HOST_TASK_EXTRA += " nativesdk-bash"
```
abelal, line 113: Not related but why do we need a space here with += ?

conf/local.conf.sample
```diff
@@ -48,275 +46,83 @@ MACHINE ??= "qemux86"
...
 
-# TOOLCHAIN_TARGET_TASK is used to add target packages to the SDK, for
-# example, to add libncurses:
+# Add target packages to the SDK
 #TOOLCHAIN_TARGET_TASK_EXTRA += " ncurses-libncurses"
```
abelal, line 116: Not related but why do we need a space here with += ?

conf/local.conf.sample
```diff
@@ -48,275 +46,83 @@ MACHINE ??= "qemux86"
...
-# User features lets you manipulate the distro features. To add a distro
-# feature, simply add it to USER_FEATURES. To remove, prefix it with ~.
-#
-# # Disable x11
-# USER_FEATURES += "~x11"
-#
-# Disable 3g by default, as most won't need it.
-USER_FEATURES += "~3g"
+# Enable and disable distro features. To disable a feature, prefix it with ~
+#USER_FEATURES += "~x11"
```
Noor-Ahsan, line 66: In my opinion we are not using it so we should remove it from here

conf/local.conf.sample
Noor-Ahsan: BTW I think customers are not using this. Should we dropt this as well?

kergoth: I think we need marketing input on this, as I think wanting to add  extra packages into an SDK would be common, such as host tooling for development, things like qmake for qt (though that particular case I'd hope it'd be automatic for images including qt6).

conf/local.conf.sample
kergoth: USER_FEATURES is used everywhere, it's our primary way of enabling and disabling features. Admittedly we have other examples of it when local.conf fragments are included for such features, but it's both commonly used for our product and is specific to us, not in the Yocto docs. I'm unsure, I think this needs discussion.

kergoth:
Action items from meeting:

- [ ] Separate issue: Remove examples from extra image features and add a reference to our ref manual
- [ ] Drop buildstats
- [ ] Strip out non-windows info from SDKMACHINE and change the default to uncomment to switch to windows
- [ ] Separate issue: remove INCOMPATIBLE_LICENSE bits

kergoth:
@Noor-Ahsan Rebased.

Noor-Ahsan:
aahhhh it need rebase again as I merged you other PR first :(

kergoth (commented):
Rebased.

PR #31, opened 2022-08-17
staging:96boards-tools: really fix resizing on the first boot

abelal:
Our earlier attempt failed due to a negligence of testing the fix with script
mode at runtime with a fresh install i.e. the change was run on an already
resized root so the issue did not show up.
This really fixes the issue by leveraging an internal option (pretend-tty-input)
of parted with the older approach of echoing and piping the confirmation to the
command.

JIRA: https://jira.alm.mentorg.com/browse/SB-19819

Signed-off-by: Awais Belal <awais_belal@mentor.com>


kergoth (approved):
LGTM via visual inspection, don't know enough about parted usage to say as to the rest, but I'm sure you've got it down.

PR #32, opened 2022-08-18
meta-sokol-flex-staging: libcoap: Introduce libcoap recipe

muhamzai:
Introduce libcoap recipe which belongs to meta-oe in
meta-networking so we can drop our maintained version
of libcoap recipe.

JIRA-ID: SB-20025


abelal (changes requested):
Rather than "Introduce libcoap recipe" we should say "backport libcoap from meta-networking". In the commit message please share the exact link from where this is being copied over.

muhamzai:
> Rather than "Introduce libcoap recipe" we should say "backport libcoap from meta-networking". In the commit message please share the exact link from where this is being copied over.

Fixed and updated.

abelal (approved):

PR #34, opened 2022-08-18
Fix issues with the MinGW SDK build

kergoth:
JIRA: SB-20197


kergoth:
A mingw32 SDK build using sokol-flex successfully builds with this applied.

abelal (approved):

PR #35, opened 2022-08-18
Add and use sdk_multilib_hook.bbclass

kergoth:
This splits out the common logic for SDK postprocessing commands which need to be run for each multilib variant, not once.


kergoth:
This class is used for any SDK postprocessing script that needs to work with or alter the environment-setup script, as those are created per-multilib, as well as those working with cb-mbs-options, for the same reason.

abelal:
@kergoth no JIRAs or short commit messages?

kergoth:
Fixed.

abelal (approved):

PR #37, opened 2022-08-23
libwebsockets: Update testapps package config

muhamzai:
Add testapps package config to disable use of testapps
in our build.

JIRA-ID: SB-20040


abelal (changes requested):
I don't think creating a package config makes sense here as our fork of libwebsockets does not allow (another discussion point IMO as I think that is incorrect) building the test-apps and us providing a package config means it is an option.
Using EXTRA_OECMAKE would be a better fit here.

muhamzai:
> I don't think creating a package config makes sense here as our fork of libwebsockets does not allow (another discussion point IMO as I think that is incorrect) building the test-apps and us providing a package config means it is an option. Using EXTRA_OECMAKE would be a better fit here.

Updated

abelal (approved):
:+1:

PR #41, opened 2022-09-01
Allow for shipment of the external toolchain with the Yocto SDK

kergoth:
JIRA: SB-19962


meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -13,8 +13,6 @@ MAINTAINER = "Siemens Digital Industries Software <embedded_support@mentor.com>"
 HOME_URL = "https://www.plm.automation.siemens.com/global/en/products/embedded/flex-os.html"
 SUPPORT_URL = "https://support.sw.siemens.com/"
 BUG_REPORT_URL = "https://support.sw.siemens.com/"
-TARGET_VENDOR = "-flex"
-SDK_VENDOR = "-flexsdk"
```
abelal, line 17: What do we get with the yocto/internal tools? Would it make sense to set these with ?= here?

abelal:
Probably squash the last two commits?

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: Both oe-core and meta-codebench-toolchain builds use the default, which is `-oe` and `-oesdk`, which I don't think is unreasonable to use even for flex builds, we don't *need* our product name in our binary prefixes.

abelal (approved):

PR #43, opened 2022-09-01
meta-sokol-flex-staging: systemd: Drop patch for ld arguments

muhamzai:
New version of systemd-boot intrduces 'combo' type for efi-ld config
in place of type 'string'. Therefore this patch is no longer needed
to fix efi-ld config from 'string' to 'array'.

JIRA-ID: SB-20059


muhamzai:
With reference to https://jira.alm.mentorg.com/browse/SB-20059

PR #44, opened 2022-09-01
Handle automatic inclusion of meta-virtualization

kergoth:
This prepares for automatic inclusion of meta-virtualization on a per-BSP basis. When it's included, the appropriate fragment will show up in `local.conf`, without having to fork the layer in question.

For BSPs which need container support, we can add an appropriate `LAYERRECOMMENDS` adjustment to our `bblayers.conf.sample` to ensure meta-virtualization is included by default on that BSP.

JIRA: SB-20067

### DIFF
--- a/meta-sokol-flex-distro/post-setup-environment
+++ b/meta-sokol-flex-distro/post-setup-environment
@@ -1,0 +1,3 @@
+if configured_layers | grep -qF meta-virtualization; then
+    load_lconf_snippet "$layer" "dynamic-layers/virtualization-layer/conf/local.conf.append"
+fi


meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
```diff
@@ -0,0 +1,3 @@
+# Uncomment the following to enable virtualization support
+#USER_FEATURES += "virtualization"
+#EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"
```
abelal, line 3: We should ideally have the second line uncommented by default so the user would only have comment/uncomment the USER_FEATURES just like we do for other features.

kergoth: Ah, good point, thanks!

kergoth: Fixed

abelal (approved):

PR #45, opened 2022-09-06
python3-unidiff: upgrade from 0.7.3 to 0.7.4

kergoth:
Upgrade from 0.7.3 to 0.7.4 to pick up
https://github.com/matiasb/python-unidiff/pull/97. This fixes the
warning encountered when running python3:do_archive_release_downloads.

JIRA: SB-20262


abelal:
@kergoth I think it would make more sense to backport the fix as a patch and have it under staging.

kergoth:
> @kergoth I think it would make more sense to backport the fix as a patch and have it under staging unless you've identified something that might not let us do that.

Good call, thanks! Please re-review.

PR #47, opened 2022-09-07
mem-per-cpu-scaling.bbclass: allow opt-out of the message

kergoth:
MEM_PER_CPU_MESSAGE may be set to 0 to disable the warning/error
message.

### DIFF
--- a/meta-sokol-flex-distro/classes/mem-per-cpu-scaling.bbclass
+++ b/meta-sokol-flex-distro/classes/mem-per-cpu-scaling.bbclass
@@ -6,7 +6,10 @@
 # PARALLEL_MAKE_SCALED with scaled down values.
 #
 # Inspired by https://elinux.org/images/d/d4/Goulart.pdf.
-#
+
+# Set to 0 to opt out of the warning/error message
+MEM_PER_CPU_MESSAGE ?= "1"
+
 # Configured using the MEM_PER_CPU_SCALING_THRESHOLDS variable:
 #
 # MEM_PER_CPU_THRESHOLD,THREADS,JOBS,ACTION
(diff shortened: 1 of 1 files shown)


kergoth:
This also updates the displayed message to show the user how to opt-out:
```
WARNING: Low memory per CPU core (~3 Gb) in this system. You may wish to reduce PARALLEL_MAKE to 16 or less in local.conf, or set MEM_PER_CPU_MESSAGE="0" to disable this messa
ge.
```

kergoth:
This also updates the displayed message to show the user how to opt-out:
```
WARNING: Low memory per CPU core (~3 Gb) in this system. You may wish to reduce PARALLEL_MAKE to 16 or less in local.conf, or set MEM_PER_CPU_MESSAGE="0" to disable this messa
ge.
```

abelal:
Hmmmmmm. I am getting a little confused here sorry. The idea that I tried to convey was to have a local.conf switch e.g. AUTOSCALE_MEM_PER_CPU. If not set we print out warnings/errors as in the usual way. If set, we auto-scale and say nothing. Would that be more user friendly?

kergoth:
[**@abelal**](https://github.com/abelal) commented on [Sep 8, 2022, 5:20 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/47#issuecomment-1240642718 "2022-09-08T12:20:54Z - Replied by Github Reply Comments"):
> Hmmmmmm. I am getting a little confused here sorry. The idea that I tried to convey was to have a local.conf switch e.g. AUTOSCALE_MEM_PER_CPU. If not set we print out warnings/errors as in the usual way. If set, we auto-scale and say nothing. Would that be more user friendly?

That's what it already does. This lets you disable the message *without* using that local.conf switch. If I know my builds will succeed with the current configuration and I don't *want* to used scaled down values, but my current ones, this lets me disable the resource check entirely.

abelal:
@kergoth thanks for the clarification.

abelal (approved):

PR #50, opened 2022-09-13
sokol-flex: correct pre-release versions

kergoth:
Replace `+snapshot-${DATE}` in `DISTRO_VERSION` with `-alpha`, and replace `-alpha` in `SDK_VERSION` with the timestamp in seconds as a suffix.

JIRA: SB-20310, SB-20454

Previous values:
```bitbake
DISTRO_VERSION="13.0+snapshot-20220913"
SDK_VERSION="13.0-snapshot"
```

New values:
```bitbake
DISTRO_VERSION="13.0-alpha"
SDK_VERSION="13.0.202209161514"
```

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -26,9 +26,11 @@ SCRIPTS_VERSION ?= "1"
 BSP_VERSION ?= "0"
 PATCH_VERSION ?= "0"
 
-SDK_VERSION := "${@'${DISTRO_VERSION}'.replace('+snapshot-${DATE}','-snapshot')}"
-DISTRO_VERSION[vardepsexclude] = "DATE"
-SDK_VERSION[vardepsexclude] = "DATE"
+SDK_VERSION = "${@d.getVar('DISTRO_VERSION').replace('+snapshot-${DATE}', '')}.${SDK_TIMESTAMP}"
+SDK_VERSION[vardepsexclude] += "SDK_TIMESTAMP"
+
+SDK_TIMESTAMP = "${@d.getVar('DATETIME')[:-2]}"
+SDK_TIMESTAMP[doc] = "The timestamp from the DATETIME variable, without seconds."
 
 # This is a single libc DISTRO, so exclude it from tmpdir name
 TCLIBCAPPEND = ""


meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -26,9 +26,11 @@ SCRIPTS_VERSION ?= "1"
...
 PATCH_VERSION ?= "0"
 
-SDK_VERSION := "${@'${DISTRO_VERSION}'.replace('+snapshot-${DATE}','-snapshot')}"
-DISTRO_VERSION[vardepsexclude] = "DATE"
-SDK_VERSION[vardepsexclude] = "DATE"
+SDK_VERSION = "${@d.getVar('DISTRO_VERSION').replace('-alpha', '')}.${DATETIME_SECS}"
+SDK_VERSION[vardepsexclude] += "DATETIME_SECS"
+
+DATETIME_SECS = "${@int(time.mktime(time.strptime('${DATETIME}', '%Y%m%d%H%M%S')))}"
+DATETIME_SECS[doc] = "The timestamp from the DATETIME variable, as seconds since 1970."
```
abelal, line 33: Hmmm isn't seconds way too granular? How about we use minutes?

kergoth: We certainly could, what's the concern with it? I'd like to note that we exclude this from the checksum, so while the timestamp will be included *when* the sdk task is run, the timestamp changing alone won't cause it to rebuild. That is, re-running your build won't build it from scratch just due to the hcange in the date, but if something else changes, it'll include the new timestamp.

kergoth: The original intention was to avoid conflicts if we rebuild, so just chose an arbitrary increasing number of the timestamp, but perhaps we should make it something actually user-comprehensible now that this isn't so specific to the ADE. I.e. YYYY.MM.DD.HH.MM or something?

abelal: I'd go with the above suggestion that you provided i.e. YYYY.MM.DD.HH.MM.

> We certainly could, what's the concern with it?

Nothing in particular I'd say. Just the seconds looked a bit too granular as I said. I don't think SDKs would be generated with such a short diff in time but minutes looked plausible.

kergoth: The more user-friendly timestamp is actually longer than the other one, but potentially less confusing. Updating.

abelal (approved):
:+1:

Noor-Ahsan:
@kergoth as we learned today that there is not alpha drop planned now for customer. I think we should update this PR accordingly and we can resolve the ticket as won't fix due to cancelation of drop to customer

Noor-Ahsan:
> @kergoth as we learned today that there is not alpha drop planned now for customer. I think we should update this PR accordingly and we can resolve the ticket as won't fix due to cancelation of drop to customer

@kergoth kindly update the PR according to this.

Noor-Ahsan:
@kergoth as we are not doing the pre release now do we need this https://github.com/MentorEmbedded/meta-sokol-flex/commit/a14ec3939d8034ecba2fb9234ea0e0071a3f817f?

kergoth:
I do think it's worthwhile during development to have some indication of how old your build is. In retrospect, however, it might be better to add 'image-buildinfo' to the image features by default in a local.font fragment instead. It's worth discussing.

Noor-Ahsan:
OK what we can do. We can merge this and then create a ticket to clean this up when we are about to be done. Thoughts?

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Sep 27, 2022, 8:35 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/50#issuecomment-1259684337 "2022-09-27T15:35:39Z - Replied by Github Reply Comments"):
> OK what we can do. We can merge this and then create a ticket to clean this up when we are about to be done. Thoughts?

Agreed.

Noor-Ahsan:
Created https://jira.alm.mentorg.com/browse/SB-20566. and merging this PR

PR #51, opened 2022-09-26
pulseaudio: make module-switch-on-connect available

akif-tariq:
This is a small change to include pusleaudio module `module-switch-on-connect` in the release.
It was noticed while testing the audio playback through g-streamer that it will be more user friendly if we add this module. This module automatically changes the sink when a new audio device is connected to host machine.


kergoth (approved):

ahsanhussain:
@akif-tariq thanks. Can you put the JIRA-ID in the commit message, and put "pulseaudio" in its current place.

PR #54, opened 2022-09-29
virtualization-layer: enable virtualization feature by default

ahsanhussain:
JIRA-ID: SB-20096

### DIFF
--- a/meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
+++ b/meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
@@ -1,3 +1,3 @@
-# Uncomment the following to enable virtualization support
-#USER_FEATURES += "virtualization"
+# Comment out the following to disable virtualization support
+USER_FEATURES += "virtualization"
 EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"


abelal:
> @Noor-Ahsan @abelal @ahsanhussain Is virtualization a BSP feature or a distro feature? This makes it enable the feature by default, but the inclusion of the layer in bblayers is a different question, so needs to be in a LAYERRECOMMENDS, either in each bsp layer, or for sokol-flex.

@kergoth I think it falls in the category of distro features however it is supported on specific BSPs (32bit not supported). Wouldn't adding the layer using '-l' won't work anymore?

kergoth:
[**@abelal**](https://github.com/abelal) commented on [Oct 2, 2022, 10:33 PM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/54#issuecomment-1264957649 "2022-10-03T05:33:46Z - Replied by Github Reply Comments"):
> > @Noor-Ahsan @abelal @ahsanhussain Is virtualization a BSP feature or a distro feature? This makes it enable the feature by default, but the inclusion of the layer in bblayers is a different question, so needs to be in a LAYERRECOMMENDS, either in each bsp layer, or for sokol-flex.
> 
> @kergoth I think it falls in the category of distro features however it is supported on specific BSPs (32bit not supported). Wouldn't adding the layer using '-l' won't work anymore?

-l works fine, I do think we should decide if we want to require that in the future, though, as I think the usability of uncommenting the USER_FEATURES in local.conf is superior to adding extra setup-environment arguments.

PR #55, opened 2022-09-29
Support on-target development for development-image, and other fixes

ahsanhussain:
This PR adds 'tools-sdk' to 'IMAGE_FEATURES_DEVELOPMENT', so that we get support for on-target development.

While at it, also add hwcodecs image feature when alsa is in COMBINED_FEATURES, and remove FEATURE_PACKAGES_hwcodecs from x11 FEATURE_PACKAGES

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -24,7 +24,7 @@ MACHINE ??= "qemux86"
 EXTRA_IMAGE_FEATURES ?= "multilib-runtime"
 
 # Image features for development-image
-IMAGE_FEATURES_DEVELOPMENT ?= "debug-tweaks"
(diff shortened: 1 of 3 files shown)


abelal (approved):

kergoth:
I have two concerns here:
1. Is hwcodecs really only *alsa* codecs? I don't think that's the case.
2. I think this leaves a gap in our provided images. development-image is now quite large and will take some time to build, but production-image doesn't allow root login and the like by default, so isn't appropriate for actual day to day *system* development. For engineers like *us*, who aren't doing on-target development, there's no good image to use for baseline testing unless we locally revert this change, or start using core-image-base or core-image-minimal-dev or something for tests.

PR #58, opened 2022-10-07
sokol-flex: default VENDOR based on MACHINE layer name

kergoth:
Determine the default vendor config filename based upon the layer name
of the layer which provides the MACHINE .conf, with meta- removed. For
example, for icicle-kit-es, `conf/vendor/polarfire-soc-yocto-bsp.conf`
would be included by default. Either VENDOR or VENDOR_CONF may be
overridden by a MACHINE, as will be the case for our official Sokol Flex
BSP machines.

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/include/vendor.inc
+++ b/meta-sokol-flex-distro/conf/distro/include/vendor.inc
@@ -1,0 +1,34 @@
+inherit layerdirs
+
+def layerdirs(d):
+    dirs = {}
+    for layerpath in d.getVar('BBLAYERS', True).split():
+        layerconf = os.path.join(layerpath, 'conf', 'layer.conf')
+
+        l = bb.data.init()
+        l.setVar('LAYERDIR', layerpath)
+        l = bb.parse.handle(layerconf, l)
+        l.expandVarref('LAYERDIR')
+
+        for layername in (l.getVar('BBFILE_COLLECTIONS', True) or '').split():
+            dirs[layername] = layerpath
+    return dirs
+
+def machine_layername(d):
+    machine_conf = bb.utils.which(d.getVar('BBPATH'), 'conf/machine/%s.conf' % d.getVar('MACHINE'))
+    if not machine_conf:
+        bb.warn('Failed to locate conf/machine/%s.conf in BBPATH' % d.getVar('MACHINE'))
+    else:
+        conf_layer = os.path.dirname(os.path.dirname(os.path.dirname(machine_conf)))
+
+        dirs = layerdirs(d)
+        for layername in (d.getVar('BBFILE_COLLECTIONS', True) or '').split():
+            layerdir = dirs.get(layername)
+            if conf_layer == layerdir:
+                return layername.replace('meta-', '')
+
+MACHINE_LAYERNAME := "${@machine_layername(d) or 'UNKNOWN'}"
+VENDOR ?= "${MACHINE_LAYERNAME}"
+VENDOR_CONF ?= "conf/vendor/${VENDOR}.conf"
+
+include ${VENDOR_CONF}
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -3,8 +3,7 @@
 # ---------------------------------------------------------------------------------------------------------------------
 
 ## Vendor Integration {{{1
-VENDOR_CONF = "${@'conf/vendor/%s.conf' % d.getVar('VENDOR') if d.getVar('VENDOR') else ''}"
-require ${VENDOR_CONF}
+require conf/distro/include/vendor.inc
 ## }}}1
 ## Sokol Flex OS Base Configuration {{{1
 DISTRO_NAME = "Sokol Flex OS"


kergoth:
This is the initial version of what we talked about last night my time, which would be used in concert with the dynamic-layers combined integration layer.

abelal:
> This is the initial version of what we talked about last night my time, which would be used in concert with the dynamic-layers combined integration layer.
> 
> This will get a different result for flex BSPs, since the layer name would be our flex ref bsps layer, so we need to ensure the flex bsps directly include this file, or the files the vendor file includes, so both flex and non-flex bsps get common integration applied.

Hmmmm. IIUC we'll have to create a conf for each of our machines which would require the conf present in the vendor integration layer for the particular vendor, right?

abelal:
> [**@abelal**](https://github.com/abelal) commented on [Oct 10, 2022, 12:37 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1272897764):
> 
> > > This is the initial version of what we talked about last night my time, which would be used in concert with the dynamic-layers combined integration layer.
> > > This will get a different result for flex BSPs, since the layer name would be our flex ref bsps layer, so we need to ensure the flex bsps directly include this file, or the files the vendor file includes, so both flex and non-flex bsps get common integration applied.
> > 
> > 
> > Hmmmm. IIUC we'll have to create a conf for each of our machines which would require the conf present in the vendor integration layer for the particular vendor, right?
> 
> I was thinking that the point of VENDOR_CONF is to get a .inc included, we could just as easily include it directly from the machine .conf rather than using the indirection of sokol-flex doing it, for the `-flex` machines. But if we think VENDOR_CONF is cleaner or more consistent, using ?= would be easy enough.

I see it as a cleaner/consistent approach. Also, the sokol-flex conf _requires_  the VENDOR_CONF so that would make sense to easily handle even for the -flex machines IMO i.e. create one and use for all (for the specific vendor).

kergoth:
> > [**@abelal**](https://github.com/abelal) commented on [Oct 10, 2022, 12:37 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1272897764):
> > > > This is the initial version of what we talked about last night my time, which would be used in concert with the dynamic-layers combined integration layer.
> > > > This will get a different result for flex BSPs, since the layer name would be our flex ref bsps layer, so we need to ensure the flex bsps directly include this file, or the files the vendor file includes, so both flex and non-flex bsps get common integration applied.
> > > 
> > > 
> > > Hmmmm. IIUC we'll have to create a conf for each of our machines which would require the conf present in the vendor integration layer for the particular vendor, right?
> > 
> > 
> > I was thinking that the point of VENDOR_CONF is to get a .inc included, we could just as easily include it directly from the machine .conf rather than using the indirection of sokol-flex doing it, for the `-flex` machines. But if we think VENDOR_CONF is cleaner or more consistent, using ?= would be easy enough.
> 
> I see it as a cleaner/consistent approach. Also, the sokol-flex conf _requires_ the VENDOR_CONF so that would make sense to easily handle even for the -flex machines IMO i.e. create one and use for all (for the specific vendor).

It uses include, not require, so I'm not clear on your logic there. Making it consistent does make sense, however. I'll update the PR.

abelal:
> > > [**@abelal**](https://github.com/abelal) commented on [Oct 10, 2022, 12:37 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1272897764):
> > > > > This is the initial version of what we talked about last night my time, which would be used in concert with the dynamic-layers combined integration layer.
> > > > > This will get a different result for flex BSPs, since the layer name would be our flex ref bsps layer, so we need to ensure the flex bsps directly include this file, or the files the vendor file includes, so both flex and non-flex bsps get common integration applied.
> > > > 
> > > > 
> > > > Hmmmm. IIUC we'll have to create a conf for each of our machines which would require the conf present in the vendor integration layer for the particular vendor, right?
> > > 
> > > 
> > > I was thinking that the point of VENDOR_CONF is to get a .inc included, we could just as easily include it directly from the machine .conf rather than using the indirection of sokol-flex doing it, for the `-flex` machines. But if we think VENDOR_CONF is cleaner or more consistent, using ?= would be easy enough.
> > 
> > 
> > I see it as a cleaner/consistent approach. Also, the sokol-flex conf _requires_ the VENDOR_CONF so that would make sense to easily handle even for the -flex machines IMO i.e. create one and use for all (for the specific vendor).
> 
> It uses include, not require, so I'm not clear on your logic there. Making it consistent does make sense, however. I'll update the PR.

My bad, misread that but it would still make sense to be able to override the VENDOR_CONF if need be.

kergoth:
Updated to:
1. keep VENDOR, define with ?=
2. define VENDOR_CONF with ?=
3. adjust the commit message to match

ahsanhussain (approved):

ahsanhussain:
Seems good but I figure we can't merge till we have done the corresponding changes for the current BSPs

abelal (approved):

abelal:
This should be tagged with a JIRA ID. @ahsanhussain which one is it for vendor alignment?

ahsanhussain:
> [**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Oct 12, 2022, 5:10 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1276056767):
> 
> > Seems good but I figure we can't merge till we have done the corresponding changes for the current BSPs
> 
> I don't believe we're actually _using_ the `VENDOR` capability anywhere yet, so I don't think it'd actually do anything to current builds.

https://github.com/MentorEmbedded/meta-flex-vendor-integration/blob/main/meta-flex-imx-integration/conf/local.conf.append#L2

ahsanhussain:
> This should be tagged with a JIRA ID. @ahsanhussain which one is it for vendor alignment?

SB-19826

kergoth:
[**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Oct 13, 2022, 2:23 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1277302733 "2022-10-13T09:23:09Z - Replied by Github Reply Comments"):
> > [**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Oct 12, 2022, 5:10 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1276056767):
> > 
> > > Seems good but I figure we can't merge till we have done the corresponding changes for the current BSPs
> > 
> > I don't believe we're actually _using_ the `VENDOR` capability anywhere yet, so I don't think it'd actually do anything to current builds.
> 
> https://github.com/MentorEmbedded/meta-flex-vendor-integration/blob/main/meta-flex-imx-integration/conf/local.conf.append#L2

This won't break with this PR, though. This PR doesn't remove VENDOR now, just changes it to default to the name based on the machine layer, so this will work fine as is.

kergoth:
[**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Oct 13, 2022, 2:25 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/58#issuecomment-1277306116 "2022-10-13T09:25:50Z - Replied by Github Reply Comments"):
> > This should be tagged with a JIRA ID. @ahsanhussain which one is it for vendor alignment?
> 
> SB-19826

Thanks, I'll add that, I hadn't realized we even had an issue to use for this :)

PR #59, opened 2022-10-10
hwcodecs: make hwcodecs part of the multimedia instead of alsa.

akif-tariq:


kergoth:
I think we should look at how hwcodecs is used in varoius bsps and bsp layers. Is it never used to pull in alsa plugins? Or should we include it for both features?

akif-tariq:
> ould look at how hwco



> I think we should look at how hwcodecs is used in varoius bsps and bsp layers. Is it never used to pull in alsa plugins? Or should we include it for both features?

@kergoth references of hwcodecs can be found in oe-core (graphics and sato recipes), inhertied in meta-imx (fsl recipes), and then again in meta-sokol-flex (graphics).

ahsanhussain:
Vendors are adding this feature to their images, but sometimes they don't make use of feature packages brought in by these. The implementation is such that hwcodecs feature can bring in packages defined by MACHINE_HWCODECS defined from the machine. Therefore we're OK with having it in multimedia USER_FEATURES but not sure whether there's a hwcodecs machine feature to warrant use of COMBINED_FEATURES.

meta-sokol-flex-support/recipes-core/images/flex-image.inc
```diff
@@ -13,6 +13,7 @@ LICENSE = "MIT"
 
 inherit core-image
 
-IMAGE_FEATURES:append:sokol-flex = " splash ${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' hwcodecs tools-audio', '', d)}"
+IMAGE_FEATURES:append:sokol-flex = " splash ${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' tools-audio', '', d)}"
+IMAGE_FEATURES:append:sokol-flex = " splash ${@bb.utils.contains('USER_FEATURES', 'multimedia', ' hwcodecs', '', d)}"
```
kergoth, line 17: multimedia is a distro feature, we should be checking `DISTRO_FEATURES` directly here.

kergoth (changes requested):

meta-sokol-flex-support/recipes-core/images/flex-image.inc
ahsanhussain: Isn't it an IMAGE_FEATURE https://github.com/MentorEmbedded/meta-sokol-flex/blob/main/conf/local.conf.sample#L43

Even if it were distro feature IIRC we've had some issues with checking USER_FEATURES added distro features inside the distro conf, hence this: https://github.com/MentorEmbedded/mel-meta-mentor-security/blame/fir_async2/meta-mentor-integrity/conf/distro/include/mel-security-ima-evm.conf#L22

akif-tariq: Ok @ahsanhussain, I have made the change as you recommended.

Appending an IMAGE_FEATURES to an IMAGE_FEATURE started giving me error:
RecursionError: maximum recursion depth exceeded while calling a Python object.

So I used IMAGE_INSTALL instead.

ahsanhussain:
@abelal @kergoth I can't think of a better way than IMAGE_INSTALL here against our multimedia and graphics IMAGE_FEATURES.

kergoth (approved):
LGTM

meta-sokol-flex-support/recipes-core/images/flex-image.inc
```diff
@@ -13,6 +13,7 @@ LICENSE = "MIT"
 
 inherit core-image
 
-IMAGE_FEATURES:append:sokol-flex = " splash ${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' hwcodecs tools-audio', '', d)}"
+IMAGE_FEATURES:append:sokol-flex = " splash ${@bb.utils.contains('COMBINED_FEATURES', 'alsa', ' tools-audio', '', d)}"
 IMAGE_INSTALL:append:sokol-flex = " util-linux-mkfs connman"
 IMAGE_INSTALL:append:sokol-flex = " ${@bb.utils.contains('BBFILE_COLLECTIONS', 'openembedded-layer', 'haveged' if not any_incompatible(d, ['haveged'], 'GPL-3.0-only') else '', '', d)}"
+IMAGE_INSTALL:append:sokol-flex = " ${@bb.utils.contains_any('IMAGE_FEATURES', ['multimedia', 'graphics'], ' ${MACHINE_HWCODECS}', '', d)}"
```
abelal, line 19: We don't need a space here before MACHINE_HWCODECS we've the needed space at the start of bb.utils.contains already.

akif-tariq: Ok thanks @abelal I have removed the space.

PR #61, opened 2022-10-12
Image Recipes: cleaned, removed and moved some of the recipes in flex layers

aouahm0d:
Following recipes have been removed:
weston-image.bb
core-image-sato.bbappend
console-image.bb

flex-initramfs-image.bb: core image recipe from distro layer moved to support layer

flex-image.inc: removed splash screen

local.conf.sample: added splash screen so that user can turn it on/off

JIRA-ID: SB-20302

Signed-off-by: Aoun Ahmed <aoun_ahmad@mentor.com>


ahsanhussain (approved):
Looks fine to me. Lets have a quick review from @Noor-Ahsan and @kergoth

kergoth (approved):
Looks reasonable to me.

PR #63, opened 2022-10-18
sokol-flex.conf: Remove old version of Ubuntu

akif-tariq:
We need to remove Ubuntu 20.04 from SANITY_TESTED_DISTROS because it is no longer a supported version as flex build host.
JIRA: SB-19958
Signed-off-by: Akif akif_tariq@mentor.com

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -67,7 +67,6 @@ PACKAGE_CLASSES ?= "package_ipk"
 
 # Sokol Flex OS's supported hosts
 SANITY_TESTED_DISTROS = "\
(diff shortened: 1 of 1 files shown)


kergoth (changes requested):
Thanks for your work on this, but we'll need to tweak the commit message some.

SB-20045 isn't the correct issue for this, this PR isn't adding support for 22.04, this PR is removing support for 20.04, which while it aligns with MELMR-904, it should probably be set to SB-19958, or add a new subtask for the removal of 20.04. Also, the description of SB-19958 still lists 20.04, which doesn't align with the MELMR anymore.

The correct syntax to refer to an issue in a commit message is: `JIRA: SB-19958` on a single line. We'll also need the commit message summary fixed, it should be prefixed based on the filename, not the path, for example: `sokol-flex: do something`.

Thanks again!

PR #64, opened 2022-10-19
docker-moby: backport 2d0f725 from newer meta-virtualization

kergoth:
This fixes builds with riscv64 by obeying `GO_LINKSHARED` from the go bbclasses.

JIRA: SB-20713

Signed-off-by: Christopher Larson <chris_larson@mentor.com>


Noor-Ahsan:
Will it has any impact on i.MX8 BSP?

kergoth:
It'll be applied generally, but it only will change behavior for systems that override go linkshared behavior, so it has no effect on arm or arm64 systems.

PR #65, opened 2022-10-20
toolchain-scripts-mingw32.bbclass: Remove trailing slash in path

muhamzai:
Modify toolchain-scripts-mingw32.bbclass to add a check in environment-setup script which removes trailing slash in path of SDKROOT.
This is needed to avoid multiple adjacent slashes in paths which are produced by appending to SDKROOT.

JIRA-ID: SB-20623


kergoth:
@muhamzai I'd suggest splitting this into two commits, one with the original file from its source, with the commit message indicating the commit hash it was pulled from, then a second with our changes to it. This makes it easier to see those changes when looking at git history, and also makes it easier to apply those changes on top of a newer upstream commit in the future if it's necessary.

abelal (approved):
👍

PR #66, opened 2022-10-21
image-root-password: use encrypted password

ahsanhussain:
The reason for this is that in Yocto Honister shadow version was updated, and shadow custom support in OE core for cleartext passwords was removed. For more details see:

https://docs.yoctoproject.org/migration-guides/migration-3.4.html#miscellaneous and
https://docs.yoctoproject.org/ref-manual/variables.html#term-EXTRA_USERS_PARAMS

JIRA-ID: SB-20424

### DIFF
--- a/meta-sokol-flex-common/classes/image-root-password.bbclass
+++ b/meta-sokol-flex-common/classes/image-root-password.bbclass
@@ -11,6 +11,8 @@
 # user will have no way to login.
 
 ROOT_PASSWORD ?= ""
+ROOT_PASSWORD_HASH_ROUNDS ?= "10000"
+
 IMAGE_INCOMPATIBLE_ZAPPED_MESSAGE = "ROOT_PASSWORD has not been set, and this \
 image has neither debug-tweaks nor empty-root-password set. This will result \
 in an image whose root login is disabled. Please set ROOT_PASSWORD to the root \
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
@abelal @kergoth kindly review.

meta-sokol-flex-common/classes/image-root-password.bbclass
```diff
@@ -20,13 +22,29 @@ allow root login with an empty password."
...
+    rounds = d.getVar('ROOT_PASSWORD_HASH_ROUNDS')
+    if rounds is not None:
+        rounds = max(1000, min(999999999, int(rounds) or 10000))
+        prefix += 'rounds={0}$'.format(rounds)
+
+    return crypt.crypt(password, prefix + salt.decode('UTF-8')).replace('$','\$')
+
+inherit extrausers
+ACTUAL_ROOT_PASSWORD = '${@encrypt_root_pw(d)}'
+EXTRA_USERS_PARAMS:prepend = "${@'usermod -p \'${ACTUAL_ROOT_PASSWORD}\' root;' if d.getVar('ACTUAL_ROOT_PASSWORD') != '' else ''}"
```
kergoth, line 47: As far as I know, his is going to expand ACTUAL_ROOT_PASSWORD regardless of the logic, due to how variable expansion happens, though it won't be *used* in all cases. Still, no point running `encrypt_root_pw` if we don't have to. You should fix this to be `'usermod -p \'%s\' root;' % d.getVar('ACTUAL_ROOT_PASSWORD')`.

kergoth:
Other than the minor quibble I commented about, this looks good to me, thanks for your work on this, it's appreciated!

meta-sokol-flex-common/classes/image-root-password.bbclass
kergoth: What I'm suggesting is this:
```
def encrypt_root_pw(d):
    import os
    import crypt
    import base64
    import shlex

    prefix = '$6$'

    password = d.getVar('ROOT_PASSWORD')
    if any(password == pattern for pattern in [ 'None', '0', '*' ]):
        return ''

    salt = base64.b64encode(os.urandom(8))

    rounds = d.getVar('ROOT_PASSWORD_HASH_ROUNDS')
    if rounds is not None:
        rounds = max(1000, min(999999999, int(rounds) or 10000))
        prefix += 'rounds={0}$'.format(rounds)

    return shlex.quote(crypt.crypt(password, prefix + salt.decode('UTF-8')).replace('$','\$'))

inherit extrausers
ACTUAL_ROOT_PASSWORD = '${@encrypt_root_pw(d)}'
EXTRA_USERS_PARAMS:prepend = "${@'usermod -p %s root;' % d.getVar('ACTUAL_ROOT_PASSWORD') if d.getVar('ACTUAL_ROOT_PASSWORD') != '' else ''}"
```

ahsanhussain: Yeah, I was confused for the usermod line; from the previous comment it looked like that an empty string case won't be handled.

ahsanhussain:
Updated and checked the shadow file generated from this; looks good. Updated the PR

ahsanhussain:
@kergoth we're importing additional python modules. For a recipe we could add to DEPENDS. How are dependencies resolved on a global scope?

kergoth:
The python modules we're importing are all in the python standard library, so are expected to already exist in the python which is running bitbake.

kergoth (approved):
LGTM

PR #67, opened 2022-10-21
sokol-flex.conf: Remove old version of Ubuntu

akif-tariq:
We need to remove Ubuntu 20.04 from SANITY_TESTED_DISTROS because it is no longer a supported version as flex build host.
JIRA: SB-20045.
Signed-off-by: Akif akif_tariq@mentor.com


akif-tariq:
Hi @Noor-Ahsan I could not change the previous PR because of my fork being deleted. I had to create a new PR.

I have also taken care of the feedback from @kergoth here: https://github.com/MentorEmbedded/meta-sokol-flex/pull/63/#pullrequestreview-1146437201

akif-tariq:
@kergoth I have to keep the JIRA issue id same as before. @Noor-Ahsan has made changes to SB-20045.

PR #69, opened 2022-10-26
flex-image: remove entropy provider packages

ahsanhussain:
We have decided to remove rng-tools and haveged like packages from our RFS on account of the fact that their use is controversial, they peg the CPU at startup, and most of all aren't needed anymore with modern SoCs which have dedicated RNG IP

JIRA-ID: SB-20711


kergoth (approved):
LGTM!

PR #70, opened 2022-10-27
sokol-flex: clean up unnecessary items

kergoth:
Clean up various unnecessary items from our DISTRO.

JIRA: SB-20618


kergoth:
This still has two open questions:

- `do_image_complete[recrdeptask] += "do_populate_lic"` is this still needed?
- Is `IMAGE_ROOTFS_EXTRA_SPACE` necessary when most builds use SD/MMC boot which resizes on first boot? It's also set to 0 in bitbake.conf and many other images.

I'm looking into the first of these now. The latter requires discussion.

abelal:
I think we're ok to drop IMAGE_ROOTFS_EXTRA_SPACE

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -192,10 +192,6 @@ INHERIT += "buildhistory"
 # Copy the image license manifest into DEPLOY_DIR_IMAGE
 INHERIT += "deploy-license-manifest"
 
-# Warn if there are available update layers which are not included in the
-# current configuration
-INHERIT += "flex-updates-check"
```
abelal, line 207: Aren't we checking the inclusion of meta-flex-cve using this class?

meta-sokol-flex-common/classes/flex-updates-check.bbclass
```diff
@@ -1,34 +0,0 @@
-# ---------------------------------------------------------------------------------------------------------------------
```
abelal, line 1: relates to the comment on the previous commit.

abelal:
Apart from one observation against flex-updates-check this is good to go.

ahsanhussain:
Can we also remove UBI_VOLNAME, APPEND and friends because we take care of such things differently in the BSP layer.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: Erk, yes, thanks, well spotted :)

kergoth:
[**@ahsanhussain**](https://github.com/ahsanhussain) commented on [Nov 1, 2022, 7:17 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/70#issuecomment-1298578310 "2022-11-01T14:17:30Z - Replied by Github Reply Comments"):
> Can we also remove UBI_VOLNAME, APPEND and friends because we take care of such things differently in the BSP layer.

The default UBI_VOLNAME includes the machine name, which we felt was redundant, it's not needed in filenames as DEPLOY_DIR_IMAGE includes MACHINE, and it's not needed on target, where we already know what we're booting, so this seems like a general fix more than machine-specifics.

The APPEND bits set in the distro are the bits of kernel configuration which aren't solely the realm of the BSP, such as splash. While we aren't obeying it in our bsps currently, I'd rather see us start doing so than scratch it entirely.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: Fixed

abelal (approved):

ahsanhussain:
We're already doing some of the bootargs related stuff corresponding to our features in flex-distro.h implementation. @abelal can we used APPEND bits in parallel, or does it make sense to use only one of these?

abelal:
> We're already doing some of the bootargs related stuff corresponding to our features in flex-distro.h implementation but that's only for U-Boot. @abelal can we used APPEND bits in parallel? Should we do it in the BSP layer like flex-distro.h?

AFAIK the APPEND syntax is only applicable to grub-efi based configurations so @ahsanhussain has a point.

ahsanhussain (approved):
After some discussion with @abelal I've created a separate improvements task for APPEND mechanism: https://jira.alm.mentorg.com/browse/SB-20844

This is good to merge.

PR #77, opened 2022-11-09
Add pyrex support

kergoth:
This is intended not for the customer as much as for internal engineers to more easily run their Linux bitbake builds in docker containers. With this:

```
git clone https://github.com/garmin/pyrex
cd build
. ./setup-environment
```

Then all your bitbake commands will automatically be run in a docker container, not natively, making it largely transparent, and enabling use of your existing tooling on the host otherwise.

### DIFF
--- a/scripts/build-setup-environment.in
+++ b/scripts/build-setup-environment.in
@@ -16,6 +16,18 @@ if [ -z "$OEINIT" ]; then
     echo >&2 "oe-init-build-env not found in @OEROOT@"
     false
 else
+    if [ -d "@OEROOT@/../pyrex" ]; then
+        echo >&2 "Using pyrex to run builds in a docker container."
+
+        PYREX_CONFIG_BIND="@OEROOT@/.."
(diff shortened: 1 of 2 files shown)


abelal:
@kergoth looks reasonable. How do we specify the container that is to be used for the build?

kergoth:
> @kergoth looks reasonable. How do we specify the container that is to be used for the build?

Good question! Pyrex has a pyrex.ini file that governs behavior. You can specify one yourself via PYREXCONFFILE, otherwise the setup script will create one in conf/ next to your local.conf and all. That file has a commented line to control the docker image to be used.

The one downside to this approach is that you *must* use pyrex's docker images, or modify your own to meet its requirements, as it has supplementary scripts it uses to set everything up. A random docker image isn't going to do the trick.

Of course, using a random docker image would have its own issues due to inconsistency in how you align the host user id and group id with the container's. Some images require they be set at docker image build time, but others do it at first login via another means, either passed environment variables which modify the running user, or via a wrapper script such as how crops does it.

PR #79, opened 2022-11-22
sokol-flex-distro: improve layer by separating user and image features 

aouahm0d:
IMAGE_FEATURES in local.conf.append files are removed to distro configuration to separate
them from USER_FEATURES

local.conf.append: removal of EXTRA_IMAGE_FEATURES from feature layer

sokol-flex.conf: Addition of EXTRA_IMAGE_FEATURES to distro layer, inclusion of distro configuration path for iot feature

JIRA: SB-20851

Signed-off-by: Aoun Ahmed <aoun_ahmad@mentor.com>

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -339,6 +339,9 @@ include conf/distro/include/flex-security.conf
 # SWUpdate configuration.
 include conf/distro/include/flex-swupdate.conf
 
+# iot-lite configuration.
+include conf/distro/include/flex-iot-lite.conf
+
 # MCF configuration
 include conf/distro/include/flex-mcf.conf
 
@@ -347,3 +350,6 @@ INITRAMFS_IMAGE ?= "flex-initramfs-image"
 INITRAMFS_IMAGE_BUNDLE ?= "${@bb.utils.contains('KERNEL_IMAGETYPES', 'fitImage', '', '1', d)}"
 ## }}}1
 # vim: set fdm=marker fdl=0 :
+
+# We have it here because we use upstream virtualization layer without any changes
+EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"
--- a/meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
+++ b/meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
@@ -1,3 +1,2 @@
 # Comment out the following to disable virtualization support
 USER_FEATURES += "virtualization"
-EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"


ahsanhussain:
@aouahm0d I had a discussion with @abelal. There are some points which we need to take care of:

- Move EXTRA_IMAGE_FEATURE:append for features other than virtualization in their respective distro include files. Create one for IOT feature because it doesn't already have so
- Keep EXTRA_IMAGE_FEATURE:append for virtualization feature here, but add a comment like "We have it here because we use upstream virtualization layer without any changes"
- You need to validate that feature overrides are applying from the distro conf; IIRC it doesn't so we have to check the DISTRO_FEATURES/USER_FEATURES directly.
- Get rid of the extra newlines

aouahm0d:
> @aouahm0d I had a discussion with @abelal. There are some points which we need to take care of:
> 
>     * Move EXTRA_IMAGE_FEATURE:append for features other than virtualization in their respective distro include files, like we discussed earlier. Create one for IOT feature because it doesn't already have one
> 
>     * Keep EXTRA_IMAGE_FEATURE:append for virtualization feature here, but add a comment like "We have it here because we use upstream virtualization layer without any changes"
> 
>     * You need to validate that feature overrides are applying from the distro conf; IIRC it doesn't so we have to check the DISTRO_FEATURES/USER_FEATURES directly.
> 
>     * Get rid of the extra newlines

Issues addressed. Validated the variables from environment as well.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -347,3 +350,6 @@ INITRAMFS_IMAGE ?= "flex-initramfs-image"
 INITRAMFS_IMAGE_BUNDLE ?= "${@bb.utils.contains('KERNEL_IMAGETYPES', 'fitImage', '', '1', d)}"
 ## }}}1
 # vim: set fdm=marker fdl=0 :
+
+# We have it here because we use upstream virtualization layer without any changes
+EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"
```
abelal, line 355: @aouahm0d first we should try if we can use feature override here. i.e.
```
EXTRA_IMAGE_FEATURES:append:feature-virtualization = " virtualization"
```
if that doesn't work we can use bb.utils.filter here.

aouahm0d: I have checked it. It does not work. So resorting to the already comitted bb.utils.filter

kergoth:
I don't think the commit message and PR summary accurately reflect what we're doing here, would you mind adjusting it? This isn't really decluttering.

kergoth:
> > I don't think the commit message and PR summary accurately reflect what we're doing here, would you mind adjusting it? This isn't really decluttering.
> 
> I was struggling to find better message. Should I rephrase it as (your suggestion): **flex: layer improvement: separate distro and user features**

Generally we stick to [OE guidelines](https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines) for our commit messages.

aouahm0d:
> > > I don't think the commit message and PR summary accurately reflect what we're doing here, would you mind adjusting it? This isn't really decluttering.
> > 
> > 
> > I was struggling to find better message. Should I rephrase it as (your suggestion): **flex: layer improvement: separate distro and user features**
> 
> Generally we stick to [OE guidelines](https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines) for our commit messages.

Yes. I acknowledge. I was only rephrasing the sentence containing the word declutter and that is what I proposed.

kergoth:
[**@aouahm0d**](https://github.com/aouahm0d) commented on [Nov 28, 2022, 10:19 PM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/79#issuecomment-1330091772 "2022-11-29T05:19:19Z - Replied by Github Reply Comments"):
> > > > I don't think the commit message and PR summary accurately reflect what we're doing here, would you mind adjusting it? This isn't really decluttering.
> > > 
> > > I was struggling to find better message. Should I rephrase it as (your suggestion): **flex: layer improvement: separate distro and user features**
> > 
> > Generally we stick to [OE guidelines](https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines) for our commit messages.
> 
> Yes. I acknowledge. I was only rephrasing the sentence containing the word declutter and that is what I proposed.

To clarify, OE guidelines mandate a certain structure to the commit messages, including prefixing the message with the filename/distro/machine and not category. That is, the prefix of `flex: layer improvement: ` doesn't fit with those guidelines, I'd expect `sokol-flex: ` to be a correct prefix for this message, but even that isn't quite correct since your commit alters multiple files. In this case it's probably trivial enough that it's alright, a larger change would require multiple commits with separate prefixes, i.e. `local.conf.append: `.

I hope I'm not coming across as critical here, your changes are good, and appreciated, I'd just like you get in the habit of following these commit message conventions as that's how we and the Yocto project do things, so it'll ease your submissions in the future. Thanks for your work on this.

ArsalanHAwan (approved):
LGTM.

Acked-by: Arsalan H. Awan <Arsalan_Awan@mentor.com>

ahsanhussain (approved):
Ideally these two changes could be distinct commits, but oh well!

PR #81, opened 2022-11-30
archive-release: archive each bundle into bundle-<hash>.tar

kergoth:
As this issue also exhibited a number of confusing warnings for the user, also quiet those.

JIRA: SB-20928


abelal (approved):

Noor-Ahsan:
@kergoth in fir we have manifest.tar file name like mel-12.0.6-imx6qsabresd-mel.manifest.tar so I think we should not rename sokol-flex-13.0+snapshot-20221122.0.0-icicle-kit-es-flex.manifest.tar to bundle-sokol-flex-13.0+snapshot-20221201.0.0-icicle-kit-es-flex.manifest.tar. More over time stamp is coming in these files names as well. we should somehow remove those time stamps otherwise layer will go in patch installer even if they not updated.

PR #85, opened 2022-12-15
flex-checkout: fix for 'sokol-flex' and +snapshot

kergoth:
Due to the use of '-' as the separator in our manifest filenames, we
have to know precisely how to get the machine name back out of it, which
means the number of dashes in the DISTRO is a factor. Now that we're
`sokol-flex` rather than `mel`, we need to adjust the cut to the 4th,
not 3rd dash, and also strip off the +snapshot- suffix.

JIRA: SB-21211

Signed-off-by: Christopher Larson <chris_larson@mentor.com>

### DIFF
--- a/scripts/release/flex-checkout
+++ b/scripts/release/flex-checkout
@@ -50,7 +50,7 @@ verify_machine () {
 set_machine () {
     workspace="$1"
     manifest_name=$(basename "$2")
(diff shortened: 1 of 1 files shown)


kergoth:
I do think it'd be good to alter things to pull the machine name out of the .info metadata rather than being based upon the manifest filename.

Noor-Ahsan:
verified by QA

PR #91, opened 2023-01-05
rebranding: removed mel occurences with flex

akif-tariq:
For re-branding requirements we need to remove all the possible occurrences of string mel or mentor with flex. Most of such files have been renamed, and occurrences of such strings.

JIRA ID: SB-21223


kergoth (approved):

kergoth:
@Noor-Ahsan ?

PR #94, opened 2023-01-09
sokol-flex-distro: relocate APPEND to flex-bsp-common

aouahm0d:
Relocate $APPEND and related variables to flex-bsp-common so that
all boot related params for uboot and grub are unified and edited
at one place.

sokol-flex-conf: removal of $APPEND and related variables so we can
shift it to flex-bsp-common.inc.

JIRA ID: SB-20844

Signed-off-by: Aoun Ahmed <aoun.ahmad.ext@siemens.com>

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -70,18 +70,6 @@ SANITY_TESTED_DISTROS = "\
     redhatenterprise*-9* \n \
 "
 
-# Sane default append for the kernel cmdline (not used by all BSPs)
-# TODO: rename APPEND to something more appropriate
-APPEND ?= "${APPEND_SERIAL} console=tty0 ${APPEND_SPLASH}"
-
-APPEND_SPLASH = "${@'quiet splash' if '${SPLASH}' else ''}"
-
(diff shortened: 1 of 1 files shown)


meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -72,7 +72,7 @@ SANITY_TESTED_DISTROS = "\
 
 # Sane default append for the kernel cmdline (not used by all BSPs)
 # TODO: rename APPEND to something more appropriate
-APPEND ?= "${APPEND_SERIAL} console=tty0 ${APPEND_SPLASH}"
+APPEND ?= "${APPEND_SERIAL} ${APPEND_SPLASH} ${mtdargs} ${APPEND_GRUB_ROOTARGS}"
```
ahsanhussain, line 75: but there are no grub specific bootargs?

ahsanhussain (changes requested):

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
ahsanhussain: Why not move all of this to flex-bsp-common.inc? And like we discussed on the chat, nothings grub specific now so you may want to rename this variable (if this is at all needed)

ahsanhussain (changes requested):

ahsanhussain (approved):

PR #95, opened 2023-01-09
Update toolchain-shar-extract.sh from oe-core bc4ee545

kergoth:
JIRA: SB-21318

### DIFF
--- a/meta-sokol-flex-staging/files/toolchain-shar-extract.sh
+++ b/meta-sokol-flex-staging/files/toolchain-shar-extract.sh
@@ -1,16 +1,16 @@
 #!/bin/sh
-
 # ---------------------------------------------------------------------------------------------------------------------
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


meta-sokol-flex-staging/files/toolchain-shar-extract.sh
```diff
@@ -1,16 +1,13 @@
 #!/bin/sh
 
-# ---------------------------------------------------------------------------------------------------------------------
-# SPDX-License-Identifier: MIT
-# ---------------------------------------------------------------------------------------------------------------------
```
ahsanhussain, line 5: Don't we still need SPDX License Identifier

kergoth: This is a tough one, we're supposed to as a matter of policy, but this also isn't a file we maintain, it's copied from upstream, so adding copyrights or license info to others files seems a bit odd to me. It's a good point, though.

kergoth: Fixed

ahsanhussain (approved):

PR #96, opened 2023-01-11
Fixes for the langdale stable branch

kergoth:
While this isn't a priority for us, keeping our public layers buildable with Yocto Project releases and stable branches is important to maintaining layer quality, and doing this will reduce the amount of work necessary when we begin our work on the next major product release.

- Base on main branch
- Merge from kirkstone
- Remove unnecessary use of `TEMPLATECONF` which now causes problems due to oe-core changes
- Update branch name from kirkstone to langdale
- Recipe version bumps


kergoth:
If you go to the https://github.com/MentorEmbedded/meta-sokol-flex/pull/96/commits page on this PR and scroll down to `Commits on Jan 3, 2023` you'll see the actual new commits below there, everything above is from the merge from kirkstone.

kergoth:
I do think our public layers need to be able to work with upstream stable branches, and we've not done a great job of this in the past.

PR #97, opened 2023-01-12
Using feature overrides to make changes, and general layer cleanup

ahsanhussain:


kergoth:
These commits are quite monolithic compared to how we usually do things. I find if you're making bullets in your commit message, it might be time to break things up. Any thoughts on this, @abelal?

ahsanhussain:
@kergoth I've reasonably split the changes. Let me know if this works. Awais will be busy, and won't be able to take a look before Monday but we have the sprint closing today.

meta-sokol-flex-staging/dynamic-layers/openembedded-layer/recipes-support/pidgin/pidgin_%.bbappend
```diff
@@ -2,4 +2,4 @@
 # SPDX-License-Identifier: GPL-2.0
 # ---------------------------------------------------------------------------------------------------------------------
 
-RDEPENDS:finch += "python3-core"
+RDEPENDS:finch:feature-sokol-flex-staging += "python3-core"
```
kergoth, line 5: This should be `:append:feature-sokol-flex-staging` to mirror existing behavior, this version is entirely replacing `RDEPENDS:finch` rather than appending to it,  unless there's a reason we're changing the behavior here.

meta-sokol-flex-staging/dynamic-layers/virtualization-layer/recipes-containers/docker-compose/python3-docker-compose_%.bbappend
```diff
@@ -2,4 +2,4 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------------------------------------------------
 
-RDEPENDS:${PN} += "${PYTHON_PN}-distutils"
+RDEPENDS:${PN}:feature-sokol-flex-staging += "${PYTHON_PN}-distutils"
```
kergoth, line 5: Same issue as finch, append is being converted to appending to an override which replaces the original rather than appending to it.

meta-sokol-flex-staging/security/recipes-ids/tripwire/tripwire_%.bbappend
```diff
@@ -3,7 +3,7 @@
 # ---------------------------------------------------------------------------------------------------------------------
 
 # Use vi by default to avoid pulling in nano
-RDEPENDS_REMOVE = "nano"
+RDEPENDS_REMOVE:feature-sokol-flex-staging = "nano"
 RDEPENDS:${PN}:remove:feature-sokol-flex-staging = " ${RDEPENDS_REMOVE}"
```
kergoth, line 7: This isn't needed, we already use this override here, right below that line, where we actually use this variable.

meta-sokol-flex-distro/recipes-core/initrdscripts/initramfs-framework_%.bbappend
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------------------------------------------------
 
-FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+FILESEXTRAPATHS:prepend:sokol-flex := "${THISDIR}/${PN}:"
 
-RDEPENDS:initramfs-module-lvm += "lvm2"
-RRECOMMENDS:${PN}-base += "initramfs-module-lvm"
+RDEPENDS:initramfs-module-lvm:sokol-flex += "lvm2"
+RRECOMMENDS:${PN}-base:sokol-flex += "initramfs-module-lvm"
```
kergoth, line 8: Same issue as finch, append is being converted to appending to an override which replaces the original rather than appending to it.

meta-sokol-flex-common/classes/mirrors.bbclass
```diff
@@ -2,11 +2,11 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------------------------------------------------
 
-PREMIRRORS:append = " \
+PREMIRRORS:append:sokol-flex = " \
 git://sources.redhat.com/ git://sourceware.org/ \n\
 "
 
-MIRRORS += "\
+MIRRORS:sokol-flex += "\
```
kergoth, line 9: Same issue as finch, append is being converted to appending to an override which replaces the original rather than appending to it.

kergoth (changes requested):

kergoth:
To clarify, I think the concept is sound, I'd say there's no need to keep the RFC tag, my only concerns are minor technical issues, you've done well with this.

meta-sokol-flex-staging/security/recipes-ids/tripwire/tripwire_%.bbappend
ahsanhussain: Must have been doing this late at night. * smh *

ahsanhussain:
@kergoth I fixed these finch issue instances for some recipes, after I got do_configure() failure with ccid and one other package. Looks like these ones got missed, as the build went through. Should have taken a look at the final diff again. Fixed now

meta-sokol-flex-common/classes/mirrors.bbclass
```diff
@@ -2,11 +2,11 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------------------------------------------------
 
-PREMIRRORS:append = " \
+PREMIRRORS:append:sokol-flex = " \
 git://sources.redhat.com/ git://sourceware.org/ \n\
 "
 
-MIRRORS += "\
+MIRRORS:append:sokol-flex = "\
```
kergoth, line 9: `:append` without a leading space. Sorry :)

kergoth (changes requested):

meta-sokol-flex-common/classes/mirrors.bbclass
ahsanhussain: Deliberately done; does it take white space or newline here?

kergoth (approved):

PR #100, opened 2023-01-18
Flex setup debian sb 20920

jainash:
Update setup-debian script

    lz4 and zstd packages are missing in setup-debian script so, adding these packages.

    JIRA: SB-20920


kergoth (approved):
Looks Good To Me.

PR #103, opened 2023-01-30
Add build information to our images

kergoth:
- flex-image.inc: include build info at /etc/build in our images
- sokol-flex: rearrange to give images a section of the file
- sokol-flex: add more variables to the included buildinfo


meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -246,6 +246,9 @@ SKIP_META_VIRT_SANITY_CHECK = "1"
 ## Images {{{1
 # We have it here because we use upstream virtualization layer without any changes
 EXTRA_IMAGE_FEATURES += "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', 'virtualization', '', d)}"
+
+# Additional variables to include in /etc/build in our images
+IMAGE_BUILDINFO_VARS:append = " DATETIME IMAGE_BASENAME IMAGE_FEATURES CORE_IMAGE_EXTRA_INSTALL"
```
abelal, line 251: a) I'd vote for adding MACHINE
b) so we don't need a vardepsexclude on the DATETIME or is that taken care of in the base implementation for buildinfo?

kergoth: It turns out the base implementation doesn't pull in the values of *any* of the variables into its checksums. This works out alright since we only care about updating the buildinfo if the image is being rebuilt *anyway*, so it's a capture of the state of things when the image was built.

kergoth: Done

ahsanhussain: ermm.. it's still an append?

abelal: a missing -f to git push I believe :D

kergoth: Ugh, oops :) Thanks

kergoth: Fixed, properly this time!

PR #104, opened 2023-02-10
Allow WSL_INTEROP into the metadata and tasks

kergoth:
JIRA: CB-17681


umairsair (approved):
This fixes the issue for me..

BTW there are some license debug variables as well which we ask user to set if he faces some issues.. IMO we should whitelist them as well.. like SALT_LOGGING_DIR.. there are many more variables as well which can be found at https://tiweb-wv.industrysoftware.automation.siemens.com/techops/confluence/display/SLD/SALT+MSL+Migration+Guide+-+Legacy+to+2.x

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -221,6 +221,10 @@ FEATURE_PACKAGES_multilib-runtime ?= "${@d.getVar('MULTILIB_RUNTIME_FEATURE_PACK
 DISTRO_EXTRA_RRECOMMENDS += "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', '', 'nss-myhostname', d)}"
 ## }}}1
 ## Workarounds & Overrides {{{1
+# Re-export and exclude from checksums
```
umairsair, line 224: just out of curiosity, why do we need to export and exclude it but not other variables like SALT ones?

umairsair (approved):

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
kergoth: We do, actually, just those ones are handled in meta-sourcery, since that's where our support for our external toolchains lie, and that's where we need to support salt. WSL_INTEROP is actually not toolchain specific, since it'd be needed to run any windows host executable.

kergoth: See https://github.com/MentorEmbedded/meta-sourcery/pull/245

PR #105, opened 2023-02-13
setup-ubuntu: install python-is-python3

akif-tariq:
bmaptools cannot find/eecute python. This is because python3 has replaced python in newer distributions. Installing python-is-python3 will fix this error since it creates a python symlink to python3.

JIRA-ID: SB-21427


abelal:
Why did we not follow the approach suggested by @kergoth here: https://jira.alm.mentorg.com/browse/SB-21427?focusedCommentId=883335&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-883335

akif-tariq:
> Why did we not follow the approach suggested by @kergoth here: https://jira.alm.mentorg.com/browse/SB-21427?focusedCommentId=883335&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-883335

python-is-python3 ships a symlink to point /usr/bin/python at the current default python3. The same sort of convenience has been provided in RedHat, both python and python3 command resolve to the same interpreter.

`‘python’: No such file or directory` message could be generated by another script/package too, so in my opinion this was a better solution.

abelal:
> > Why did we not follow the approach suggested by @kergoth here: https://jira.alm.mentorg.com/browse/SB-21427?focusedCommentId=883335&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-883335
> 
> python-is-python3 ships a symlink to point /usr/bin/python at the current default python3. The same sort of convenience has been provided in RedHat, both python and python3 command resolve to the same interpreter.
> 
> `‘python’: No such file or directory` message could be generated by another script/package too, so in my opinion this was a better solution.

Doing this wouldn't we be changing the user defaults if one has python set to point to python2 for whatever reason? If this is needed by the build system as a whole that would be understandable but not otherwise IMO.

@kergoth thoughts?

akif-tariq:
> > > Why did we not follow the approach suggested by @kergoth here: https://jira.alm.mentorg.com/browse/SB-21427?focusedCommentId=883335&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-883335
> > 
> > 
> > python-is-python3 ships a symlink to point /usr/bin/python at the current default python3. The same sort of convenience has been provided in RedHat, both python and python3 command resolve to the same interpreter.
> > `‘python’: No such file or directory` message could be generated by another script/package too, so in my opinion this was a better solution.
> 
> Doing this wouldn't we be changing the user defaults if one has python set to point to python2 for whatever reason? If this is needed by the build system as a whole (not just that pre-built bmap-tools) that would be understandable but not otherwise IMO.
> 
> @kergoth thoughts?
> 
> Note: there's also a new 3.6 release for bmap-tools, maybe we should consider and update as well?

@abelal Python2 has been deprecated in most distributions. (as I mentioned earlier, for Redhat both python and python3 command lines would resolve to python3 interpreter). Even then, if someone installs python2 in Ubuntu, python command will still resolve to python3 (assuming python-is-python3 is installed). To execute python2 interpreter, one will have to execute python2 from the command line.

kergoth:
We don't care what `python` points to, and we don't use python2 for anything. We should fix the `#!` on bmaptool to align with that of every other tool we use, including bitbake, and point directly at python3.

meta-sokol-flex-support/recipes-core/meta/archive-release.bb
```diff
@@ -741,6 +741,7 @@ do_archive_images () {
         echo "${WORKDIR}/xlayers.conf" >>include
     fi
 
+    sed -i 's//usr/bin/env python//usr/bin/env python3/g' "${WORKDIR}/bmaptool"
```
abelal, line 744: I'd suggest using a different symbol than / for the sed fields separator as we already have /(s) in the path. i.e.
```
sed -i 's#/usr/bin/env python#/usr/bin/env python3#g' "${WORKDIR}/bmaptool"
```

Also, please update the commit title so we know what we're modifying exactly. I'd probably go with
```
archive-release:bmap-tools: use python3 for the shebang
```
but you can pick something else if you think that's more suitable.

akif-tariq: > I'd suggest using a different symbol than / for the sed fields separator as we already have /(s) in the path. i.e.
> 
> ```
> sed -i 's#/usr/bin/env python#/usr/bin/env python3#g' "${WORKDIR}/bmaptool"
> ```
> 
> Also, please update the commit title so we know what we're modifying exactly. I'd probably go with
> 
> ```
> archive-release:bmap-tools: use python3 for the shebang
> ```
> 
> but you can pick something else if you think that's more suitable. The commit message does not tell what the problem is in this case.

Made the suggested changes.

abelal (approved):

PR #107, opened 2023-03-07
sokol-flex: gather info to diagnose licensing network issues

kergoth:
- Add the value of SOURCERY_LICENSE_NETWORK_TASKS to the Build
  Configuration header at the start of a bitbake build.
- Add debug messages to task logs with the value of the task's `network`
  flag before running the task.

Example addition to Build Configuration header:

    SOURCERY_LICENSE_NETWORK_TASKS = "    do_configure     do_compile     do_compile_kernelmodules     do_install     do_bundle_initramfs     do_configure_ptest     do_compile_ptest     do_install_ptest "

Example addition to a task log file:

    DEBUG: glibc-2.35-r0 do_configure: Environment: do_configure[network]="1"

JIRA: SB-21627

### DIFF
--- a/meta-sokol-flex-common/classes/taskenvlog.bbclass
+++ b/meta-sokol-flex-common/classes/taskenvlog.bbclass
@@ -1,0 +1,28 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+
+# Log the values of specified variables when starting tasks as debug messages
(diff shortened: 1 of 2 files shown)


meta-sokol-flex-common/classes/taskenvlog.bbclass
```diff
@@ -0,0 +1,28 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
```
umairsair, line 2: just for my knowledge, why its with MIT license?

PR #113, opened 2023-03-30
kernel-fitimage: fix the default config check

aarslan7:
The current check for default dtb image checks if the file exists and is
not empty but appends a slash to the path due to which the file is never
found. It also doesn't replace slash in filename with _ as done when
populating the DTB variable.
    
JIRA: SB-21047


aarslan7:
> As mentioned, we actually need to drop the class from here altogether.

@abelal sure, I'll remove it. i thought there was a special requirement so we were keeping this.

aarslan7:
@abelal updated

PR #114, opened 2023-04-07
setup-environment: fix argument handling for args >9

kergoth:
This script iterates through the arguments solely to get the correct `$BUILDDIR`, but doesn't have full fledged option parsing since it's actually a light wrapper around setup-flex-builddir.  During this iteration, it seems '$10' gets expanded as '$1' followed by a 0, which isn't what we're looking for here, so use '${}' syntax instead.

JIRA: SB-21796

### DIFF
--- a/setup-environment
+++ b/setup-environment
@@ -38,10 +38,10 @@ else
     fi
 
     for i in $(seq $#); do
-        setup_flex_arg="$(eval printf "%s" "\$$i")"
+        setup_flex_arg="$(eval printf "%s" "\${$i}")"
         case "$setup_flex_arg" in
             -b)
-                BUILDDIR="$(eval printf "%s" "\$$((i + 1))")"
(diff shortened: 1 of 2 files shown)


setup-environment
```diff
@@ -38,7 +38,7 @@ else
     fi
 
     for i in $(seq $#); do
-        setup_flex_arg="$(eval printf "%s" "\$$i")"
+        setup_flex_arg="$(eval printf "%s" "\${$i}")"
         case "$setup_flex_arg" in
             -b)
                 BUILDDIR="$(eval printf "%s" "\$$((i + 1))")"
```
abelal, line 44: Hmmmm don't we need the above magic here as well? Would we be better off implementing actual arg parsing?

abelal (approved):

kergoth:
@Noor-Ahsan  Should be ready to merge

PR #116, opened 2023-05-02
lvm2: add lvm2-flex.service file

akif-tariq:
lvm2 bbappend is fixed to add lvm2-flex.service file.


abelal (approved):

meta-sokol-flex-distro/dynamic-layers/openembedded-layer/recipes-support/lvm2/lvm2_%.bbappend
```diff
@@ -3,6 +3,6 @@
 # ---------------------------------------------------------------------------------------------------------------------
 
 FLEX_LVM_PKG = ""
-FLEX_LVM_PKG:sokol-flex = "recipes-support/lvm2/lvm2-flex.inc"
+FLEX_LVM_PKG:sokol-flex = "lvm2-flex.inc"
 
 include ${FLEX_LVM_PKG}
```
aarslan7, line 8: you should use require here instead of include. it will make sure the file is added when needed and we won't have such problems in the future
cc @abelal

PR #121, opened 2023-05-08
relocate_sdk.sh: find ld-linux only at toplevel

kergoth:
Also ensure we only select one.

Without this change, it may select the .debug version of ld-linux in
SDKs which include debug info, which is usual.

JIRA: SB-22047

### DIFF
--- a/meta-sokol-flex-staging/files/relocate_sdk.sh
+++ b/meta-sokol-flex-staging/files/relocate_sdk.sh
@@ -43,7 +43,7 @@ if [ -e "$native_sysroot/lib" ]; then
         exit 1
     fi
 
-    dl_path=$(find $native_sysroot/lib -name "ld-linux*")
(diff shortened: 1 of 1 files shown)


meta-sokol-flex-staging/files/relocate_sdk.sh
```diff
@@ -43,7 +43,7 @@ if [ -e "$native_sysroot/lib" ]; then
         exit 1
     fi
 
-    dl_path=$(find $native_sysroot/lib -name "ld-linux*")
+    dl_path=$(find $native_sysroot/lib -mindepth 1 -maxdepth 1 -name "ld-linux*" | head -n 1)
```
ahsanhussain, line 46: Can't we simply do "ls $native_sysroot/lib/ld-linux* | head -n 1" instead?

Noor-Ahsan: Based on Chris comment I am merging this change as Customer has tested this.

PR #122, opened 2023-05-09
setup scripts: install python3-setuptools package. 

ArsalanSaleemGhuman:
The setuptools package is required by recipetool utility.
    Explicitly installing the setuptool package in setup-rh
    script to fulfil the dependency.

    JIRA: SB-22045

    Signed-off-by: Arsalan Saleem <arsalan_saleem@mentor.com>

### DIFF
--- a/scripts/setup-rh
+++ b/scripts/setup-rh
@@ -10,7 +10,7 @@
 # kind, whether express or implied.
 
 packages="python39 \
-          python39-setuptools \
+          python3-setuptools \
           patch diffstat git bzip2 tar \
           gzip gawk chrpath wget cpio perl file which \
           make gcc gcc-c++ rsync rpcgen \


Noor-Ahsan:
@ArsalanSaleemGhuman can you also share the evidance that this package is already there or in Jira and point reviewer there

ArsalanSaleemGhuman:
@Noor-Ahsan ,
I have already added the information on JIRA ticket. See image below.
image

aarslan7 (approved):

abelal (changes requested):
The commit title should be
```
setup-rh: install/remove <package-name>
```

just saying "update" is way too generic. Also I think a more appropriate way would be to explicitly install the python3-setuptools (or whatever the package name is for RhEL) now that we know there's a dependency.

ArsalanSaleemGhuman:
> The commit title should be
> 
> ```
> setup-rh: install/remove <package-name>
> ```
> 
> just saying "update" is way too generic. Also I think a more appropriate way would be to explicitly install the python3-setuptools (or whatever the package name is for RhEL) now that we know there's a dependency.

sure let me add changes to install the package.

abelal:
@ArsalanSaleemGhuman why do we have 2 commits here now? I don't think the first one's needed. Also have you confirmed that the package name is "python3-setuptools" for RHEL?

abelal:
@ArsalanSaleemGhuman why do we have 2 commits here now? I don't think the first one's needed. Also have you confirmed that the package name is "python3-setuptools" for RHEL?

ArsalanSaleemGhuman:
@abelal  i have squashed the commits.

abelal (approved):

abelal:
> > @ArsalanSaleemGhuman why do we have 2 commits here now? I don't think the first one's needed. Also have you confirmed that the package name is "python3-setuptools" for RHEL?
> 
> Does it matter how many commits are there? Don't we squash and merge ?
> 
> Yes i have verified installing the python3-setuptools on RedHat.

@aarslan7 we merge as-is with an additional merge commit i.e. we don't squash and merge.

PR #123, opened 2023-05-18
sokol-flex.conf: add feature support for encrypted-fs

akif-tariq:
Support for encrypted-fs has been implemented for imx in meta-flex-ref-bsps repository. This change enables the encrypted-fs feature when it is enabled in main
local.conf

JIRA-ID: SB-21384

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -243,6 +243,9 @@ IMAGE_BUILDINFO_VARS ?= "\
     DATETIME \
     IMAGE_BASENAME IMAGE_FEATURES CORE_IMAGE_EXTRA_INSTALL \
 "
+
+# Encrypted FS
+EXTRA_IMAGE_FEATURES:append:feature-encrypted-fs = " encrypted-fs"
 ## }}}1
 ## SDK & Application Development Environment {{{1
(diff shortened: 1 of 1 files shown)


meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -349,3 +349,6 @@ INITRAMFS_IMAGE ?= "flex-initramfs-image"
 INITRAMFS_IMAGE_BUNDLE ?= "${@bb.utils.contains('KERNEL_IMAGETYPES', 'fitImage', '', '1', d)}"
 ## }}}1
 # vim: set fdm=marker fdl=0 :
+
+# Encrypted FS
```
Noor-Ahsan, line 353: This should not be added here. it should go to imx-common or BSP layer. See where we added this in Fir

akif-tariq: @Noor-Ahsan I made this change because of @ahsanhussain and @abelal comments here:

https://github.com/MentorEmbedded/meta-flex-ref-bsps/pull/124#discussion_r1189860939

Noor-Ahsan: OK understood

abelal: This change needs to be moved to the images section in the same file.

PR #125, opened 2023-06-05
busybox: enable base64 utility

tarmeh2r:
Qt creator IDE checks target connectivity and presence of some utilities on the target rootfs to perform device test. The test fails if base64 utility is not found. To address this, enable base64 utility by default for our distro using CONFIG_BASE64 config fragment.

JIRA-ID: SB-21059


tarmeh2r:
the commit message and config fragment still contain references of Qt Creator and I am not exactly sure if those need to be removed in the context of this layer. But please feel free to point out if further changes are required

ahsanhussain (approved):

kergoth (approved):
I think it's good that we mention creator, so it's clear *why* we're doing this. It's always helpful to include the why, not just the what, in commit messages. Thanks. LGTM

PR #128, opened 2023-06-21
lvm2-flex.service: replace udev.service with systemd-udevd.service

akif-tariq:
udev.service does not exist, so there is no point of having it in the `After` unit dependency directive. I think the suitable parameter is `systemd-udevd.service`.

JIRA-ID: SB-22306


ahsanhussain:
Please explain the intent of this change in the commit message.

abelal:
This looks correct however have we made any observations at runtime after this dependency change? i.e. did we test encrypted-fs after this change?

abelal:
Also why haven't we considered the second point mentioned in the JIRA ticket?

> ExecStop is set to /usr/sbin/pvchange -an which when executed complaints about the -n parameter which does not exist.

PR #129, opened 2023-06-30
psplash: disable startup message

vj-kumar:
By default startup message is enabled. This draws a stripe close to the bottom of the screen with the background color specified(currently black).

This might not be visible in some cases, where the image is centered and the remaining area is filled with the background color.

Disable startup message to avoid the black rectangle displayed over the psplash image.


vj-kumar:
@Noor-Ahsan Not sure if this is relevant here but we are already disabling this manually in couple of our downstream layers. 

cc: @Riteshkr22

abelal:
Is this intended for the main branch or kirkstone?

Noor-Ahsan:
> Is this intended for the main branch or kirkstone?

I think they want it on Kirkstone. @vj-kumar  kindly update the branch

vj-kumar:
> @vj-kumar kindly update the branch t kirkstone so that we can merge this

Done @Noor-Ahsan

Noor-Ahsan (approved):

PR #131, opened 2023-07-18
scripts:setup-env: set LC_CTYPE for bitbake

aarslan7:
With the recent bitbake update, it is utilizing locale.getlocale() and locale.setlocale() from the python locales module. This module has a limitation when the host doesn't have the encoding defined, so e.g. if we have:
```
   $ locale
   ...
   LC_CTYPE="en_IN"
   ...
```
it will not pick the correct encoding:
```
   $ python3
   Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
   Type "help", "copyright", "credits" or "license" for more information.
   >>> import locale
   >>> locale.getlocale()
   ('en_IN', 'ISO8859-1')
```
and it will fail:
```
  >>> default_locale = locale.getlocale(locale.LC_CTYPE)
  >>> print(default_locale)
  ('en_IN', 'ISO8859-1')
  >>> locale.setlocale(locale.LC_CTYPE, default_locale)
  Traceback (most recent call last):
    File "<stdin>", line 1, in <module>
    File "/usr/lib/python3.10/locale.py", line 620, in setlocale
      return _setlocale(category, locale)
  locale.Error: unsupported locale setting
```
By setting the locale using LC_CTYPE="en_US.UTF-8", this issue is resolved for bitbake.

JIRA-ID: SB-22332

### DIFF
--- a/scripts/build-setup-environment.in
+++ b/scripts/build-setup-environment.in
@@ -9,6 +9,7 @@ elif [ -n "$ZSH_NAME" ]; then
 else
     BUILDDIR="$(pwd -P)"
 fi
+export LC_CTYPE="en_US.UTF-8"
 export TEMPLATECONF="@TEMPLATECONF@"
 OEINIT="$(find "@OEROOT@" -maxdepth 1 -name \*-init-build-env | head -n 1)"
 if [ -z "$OEINIT" ]; then


aarslan7:
This issue is coming up because of this commit:
https://git.openembedded.org/bitbake/commit/?h=yocto-4.0.11&id=0c6f86b60cfba67c20733516957c0a654eb2b44c

aarslan7:
bitbake needs to use `locale.setlocale(locale.LC_ALL, "")` instead when setting back to default locale to fix this issue:
https://docs.python.org/3/library/locale.html#locale.setlocale

kergoth:
As far as I know, bitbake and OE specifically require python be started with a utf-8 locale variant. You're trying to use an ISO one, which is likely to fail the sanity check even if you do try to adjust bitbake behavior.

aarslan7:
@kergoth this issue is happening on all machines in banglore, india. they have the locale set as "en_IN" and no encoding so i assume that means ISO encoding. our setup scripts install the "en_US.utf-8" font and that was enough in the past since bitbake was just checking if en_US.utf-8 is available, but now it tries to switch locales in the latest update, which causes this error on banglore machines. my machines are using en_US.utf-8 as the default locale, so i didn't see this issue on my side

kergoth:
[**@aarslan7**](https://github.com/aarslan7) commented on [Jul 24, 2023, 11:55 PM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/131#issuecomment-1649233373 "2023-07-25T06:55:37Z - Replied by Github Reply Comments"):
> @kergoth this issue is happening on all machines in banglore, india. they have the locale set as "en_IN" and no encoding so i assume that means ISO encoding. our setup scripts install the "en_US.utf-8" font and that was enough in the past since bitbake was just checking if en_US.utf-8 is available, but now it tries to switch locales in the latest update, which causes this error on banglore machines. my machines are using en_US.utf-8 as the default locale, so i didn't see this issue on my side
> 
> _Originally posted by @aarslan7 in https://github.com/MentorEmbedded/meta-sokol-flex/issues/131#issuecomment-1649233373_

I don't believe it's supposed to *require* the user locale be set to en_US.UTF-8 to kick off a build, but given oe-core's bitbake.conf forces LC_ALL to it for subprocesses anyway, I don't feel too strongly about avoiding it either.

kergoth (approved):

PR #133, opened 2023-08-02
sokol-flex: fix getvar call which broke sdk builds

kergoth:
There was a typo: getvar instead of getVar, which causes the do_populate_sdk
task to fail when these variables are used in our postprocessing functions.

JIRA: SB-22499

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -261,7 +261,7 @@ SDKPATHINSTALL = "~/${DISTRO}/sdk/${SDK_VERSION}/${IMAGE_BASENAME}-${MACHINE}"
 
 # Current multilib prefix, for non-multilib images. Ex. the lib32
 # environment-setup within a non-lib32 image SDK.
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
@kergoth  can you add what issue it is fixing? just adding that fixed type does not help what was actual issues whic we were facing

kergoth:
[**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Aug 4, 2023, 1:30 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/133#issuecomment-1665220460 "2023-08-04T08:30:46Z - Replied by Github Reply Comments"):
> [@kergoth](https://github.com/kergoth) can you add what issue it is fixing? just adding that fixed type does not help what was actual issues whic we were facing

I already said that it broke the sdk builds, but I can elaborate if needed..

kergoth:
> [**@Noor-Ahsan**](https://github.com/Noor-Ahsan) commented on [Aug 4, 2023, 1:30 AM MST](https://github.com/MentorEmbedded/meta-sokol-flex/pull/133#issuecomment-1665220460):
> 
> > [@kergoth](https://github.com/kergoth) can you add what issue it is fixing? just adding that fixed type does not help what was actual issues whic we were facing
> 
> I already said that it broke the sdk builds, but I can elaborate if needed..

Done.

PR #134, opened 2023-08-10
flex-checkout: re-prompt if .manifest is a zero-byte file

kergoth:
This occurs if the manifest prompt is interrupted (ex. ^C). By checking
to see if the file exists and is non-empty when determining whether to
prompt, we can avoid this error:

    sed: can't read .info: No such file or directory

JIRA: SB-22545

### DIFF
--- a/scripts/release/flex-checkout
+++ b/scripts/release/flex-checkout
@@ -189,7 +189,7 @@ done
 
 manifest="$1"
 if [ -z "$1" ]; then
-    if ! [ -e "$project/.manifest" ]; then
+    if ! [ -s "$project/.manifest" ]; then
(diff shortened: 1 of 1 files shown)


abelal (approved):
@kergoth just for my knowledge, are we not handling ctrl-c and friends to be able to exit cleanly?

kergoth:
I don't believe any of our shell scripts are setting up signal handlers at this time. The good thing is that the scripts are intended to be able to be re-run against an existing workspace, so leaving behind the remnants isn't particularly harmful. Adding a cleanup task to remove the workspace on failure is viable, I think, but we'd have to make sure we only clean up workspaces we've created this run, since it can also be re-run against an exissting workspace, which we wouldn't want to remove. I do think that's the best path, but chose the simpler fix for the short term to reduce risk.

PR #135, opened 2023-08-11
qt6: sdk: rewrite cmake toolchain file without environment variables

kergoth:
Cherry-pick [5cbac6cf356b94db8dbadbc7b3b9149b6374ef72](https://code.qt.io/cgit/yocto/meta-qt6.git/commit/?h=5cbac6cf356b94db8dbadbc7b3b9149b6374ef72) from meta-qt6 to
fix populate_sdk failure due to libdir mismatch between nativesdk and
target.

JIRA: SB-22426

### DIFF
--- a/meta-sokol-flex-staging/classes/populate_sdk_qt6_base.bbclass
+++ b/meta-sokol-flex-staging/classes/populate_sdk_qt6_base.bbclass
@@ -1,0 +1,112 @@
+inherit qt6-paths
+
+SDK_POSTPROCESS_COMMAND:prepend = "create_qt6_sdk_files;"
+
+
+PATH_DELIM = ":"
+PATH_DELIM:sdkmingw32 = ";"
+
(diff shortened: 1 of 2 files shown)


kergoth:
Updated to link to the [commit](https://code.qt.io/cgit/yocto/meta-qt6.git/commit/?h=5cbac6cf356b94db8dbadbc7b3b9149b6374ef72) we're cherry-picking from [meta-qt6](https://code.qt.io/cgit/yocto/meta-qt6.git/) to ease review.

Noor-Ahsan:
@kergoth Bangalore team is validating this PR by testing the feature. We will merge this after their confirmation

abelal (approved):
This looks correct however as @Noor-Ahsan already pointed out I think we should test the Qt Creator workflows with this once because we have a little divergence there from upstream.

PR #139, opened 2023-08-24
Include multilib library paths in ld.so configuration

kergoth:
A new recipe installs extra ld.so configuration files for multilib
library paths, to ensure the dynamic linker and ldconfig can find all
the installed libraries.

This gives us correct ldconfig behavior, where `ldconfig -p` lists the
libraries from all multilib library paths.

Before:
```
root@qemuarm64:~# ldconfig -p|grep liblttng-ust-dl
        liblttng-ust-dl.so.1 (libc6,hard-float) => /usr/lib/liblttng-ust-dl.so.1
```

After:
```
root@qemuarm64:~# ldconfig -p|grep liblttng-ust-dl
        liblttng-ust-dl.so.1 (libc6,AArch64) => /usr/lib64/liblttng-ust-dl.so.1
        liblttng-ust-dl.so.1 (libc6,hard-float) => /usr/lib/liblttng-ust-dl.so.1
```

JIRA: SB-25565

### DIFF
--- a/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
+++ b/meta-sokol-flex-distro/conf/distro/sokol-flex.conf
@@ -231,6 +231,10 @@ do_sca_tracefiles[dirs] =+ "${SCA_SOURCES_DIR}"
 
 # Disable the meta-virtualization layer inclusion warning
 SKIP_META_VIRT_SANITY_CHECK = "1"
+
+# Add all the library paths to ld.so.conf, as the default glibc ldconfig
+# binary does not search for all the multilibs.
+DISTRO_EXTRA_RDEPENDS += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'sokol-flex-staging', multilib_pkg_extend(d, 'ldso-config'), '', d)}"
 ## }}}1
 ## Images {{{1
 # We have it here because we use upstream virtualization layer without any changes
--- a/meta-sokol-flex-staging/recipes-core/glibc/ldso-config.bb
+++ b/meta-sokol-flex-staging/recipes-core/glibc/ldso-config.bb
@@ -1,0 +1,18 @@
+SUMMARY = "ld.so configuration file for the current tuning library paths"
+DESCRIPTION = "${SUMMARY}"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/COPYING.MIT;md5=3da9cfbcb788c80a0384361b4de20420"
+INHIBIT_DEFAULT_DEPS = "1"
+
+deltask do_fetch
+deltask do_unpack
+deltask do_patch
+deltask do_configure
+deltask do_compile
+
+do_install () {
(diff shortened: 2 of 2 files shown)


meta-sokol-flex-staging/recipes-core/glibc/ldconfig-multilib.bb
```diff
@@ -0,0 +1,34 @@
...
+    from pathlib import Path
+
+    ldsodir = Path(d.getVar('D')) / d.getVar('sysconfdir')[1:] / 'ld.so.conf.d'
+    ldsodir.mkdir(parents=True, exist_ok=True)
+
+    libdir_vars = ['base_libdir', 'libdir']
+    for multilib, mldata in per_multilib(d):
+        libdirs = sorted(set(mldata.getVar(var) for var in libdir_vars))
+        defaulttune = mldata.getVar('DEFAULTTUNE')
+        fname = '{1}.conf'.format(defaulttune)
```
Noor-Ahsan, line 25: @kergoth  getting this error

ERROR: ldconfig-multilib-1.0-r0 do_install: Error executing a python function in exec_func_python() autogenerated:

The stack trace of python calls that resulted in this exception/failure was:
'exec_func_python() autogenerated', lineno: 2, function: <module>
     0001:
 *** 0002:do_install(d)
     0003:
'/var/jenkins/workspace/mel_ginkgo_async1_main_incremental/buildtype/mel/label/flex-u22/machine/imx8mpevk-flex/repotop/meta-sokol-flex/meta-sokol-flex-staging/recipes-core/glibc/ldconfig-multilib.bb', lineno: 25, function: do_install
     0021:    libdir_vars = ['base_libdir', 'libdir']
     0022:    for multilib, mldata in per_multilib(d):
     0023:        libdirs = sorted(set(mldata.getVar(var) for var in libdir_vars))
     0024:        defaulttune = mldata.getVar('DEFAULTTUNE')
 *** 0025:        fname = '{1}.conf'.format(defaulttune)
     0026:        with open(ldsodir / fname, 'w') as f:
     0027:            f.writelines(l + '\n' for l in libdirs)
     0028:}
     0029:
Exception: IndexError: Replacement index 1 out of range for positional args tuple

kergoth: Fixed, thanks! No idea how I let that slip in.

kergoth:
Just a note that this was implemented as a standalone package/recipe rather than altering glibc to simplify internal vs external toolchain, and also to avoid unnecessary rebuilds of everything depending on glibc when this is added and changed.

kergoth:
Revamped to rely on bitbake's existing multilibs handling, rather than providing a single recipe that goes over all the multilibs itself the way meta-environment does.

kergoth:
Just adding a note here for future reference: this issue needs fixing upstream in oe-core, either via default ld.so conf files like this, but integrated into the default nodistro setup, or by fixing ldconfig's defaults, which is why this new recipe resides in the staging layer rather than the distro layer, even though we probably don't want to submit it directly in this form.

PR #141, opened 2023-09-06
flex_utils/dump_licenses: use 'a+' mode with append flag

tarmeh2r:
We want to create a file if it doesn't already exist, but we also want to read it if it already exists to create a consolidated set of licenses. We get FileNotFoundError with 'r' mode if we make the first call to the function with --append. We have such a use case in jenkins_mel script when we call build_graph() first for the external toolchain and then again for the internal toolchain. Therefore, use 'a+' mode instead of 'r'.

JIRA: SB-22425


meta-sokol-flex-distro/lib/bblayers/flex_utils.py
```diff
@@ -295,7 +295,8 @@ def do_dump_licenses(self, args):
 
         # if append flag is set then read the existing file first and create a consolidated set
         if args.append:
-            with open(filename, 'r') as f:
+            with open(filename, 'a+') as f:
+                f.seek(0)
```
abelal, line 299: Hmmm checking the existence of the file would make more sense in this case, no?
```
if args.append and os.path.exists(filename):
```
thoughts?

tarmeh2r: Yes, that would be more appropriate. Thanks. Let me update the PR and do a quick validation.

abelal (approved):

PR #145, opened 2023-09-14
Disable tty1 login for BSPs that choose to do so

kergoth:
BSPs that want it disabled may add disable-tty1-login to their
MACHINE_FEATURES, which will result in its addition to IMAGE_FEATURES
for flex images, which will disable the login.

JIRA: SB-22595


kergoth:
See also https://github.com/MentorEmbedded/meta-flex-ref-bsps/pull/157.

abelal (approved):

PR #146, opened 2023-09-22
weston-init: set WAYLAND_DISPLAY in the default shell profile

kergoth:
Set WAYLAND_DISPLAY by default for console or ssh users so they are able
to run wayland programs more easily. This fixes the ability to run
programs over ssh for Qt Creator, for example.

JIRA: SB-22247

### DIFF
--- a/meta-sokol-flex-distro/recipes-graphics/wayland/weston-init.bbappend
+++ b/meta-sokol-flex-distro/recipes-graphics/wayland/weston-init.bbappend
@@ -1,0 +1,15 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


abelal (approved):
LGMT.

@kergoth what have we decided for the other part of the problem where the user needs to be in the wayland and video groups?

abelal:
> @abelal That might be true for non-root users, but root is able to do it without that, so I'm not sure it's necessary at this time. We have a root user that doesn't need it, a non-root user that does but is already in the groups, and a mechanism by which new users can be added or added to those groups (extrausers.bbclass), so I don't think we need to add another new variable to do it. That's my personal opinion, at least.

agreed, that should be good enough!

ahsanhussain (approved):

Noor-Ahsan:
@syamasundara @jainash I am merging this so kindly do sanity testing on Qt on i.mx8 series BSPs

PR #147, opened 2023-09-22
lttng-ust: upgrade from 2.13.5 to 2.13.6

kergoth:
This is necessary to fix a tracing issue where source lookup is disabled
because the"ip" context information is missing.

JIRA: SA-5426

### DIFF
--- a/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-ust_2.13.6.bb
+++ b/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-ust_2.13.6.bb
@@ -1,0 +1,15 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


abelal:
@kergoth there's a typo in the commit message
```
ource -> source
```

Would it make sense to backport the required patch (if we've identified already) for the ip specifics to work? Also, I see that upstream already has 2.13.6 for kirkstone latest, when do we plan to upgrade to that?

Noor-Ahsan:
@abelal  yocto update is planned near Dec

kergoth:
Given it's a minor version, upstream is already moving to it, and this is the recommendation by franz, I think this is the best approach.

kergoth:
Fixed the message typo, thanks!

Noor-Ahsan:
@kergoth  we are updating the version to 2.13.6 in 2.13.5 recipe. I think we should fix that?

kergoth:
@Noor-Ahsan Updated to avoid using bbappend to do a version upgrade.

PR #148, opened 2023-09-29
Fix lttng-tools multilib issues

kergoth:
* Patch lttng-tools to obey its configure arguments for consumerd paths for multilibs
* Adjust lttng-tools packaging to avoid 32-bit binaries overwriting 64-bit ones

### DIFF
--- a/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_%.bbappend
+++ b/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_%.bbappend
@@ -6,8 +6,11 @@ FILESEXTRAPATHS:prepend:feature-sokol-flex-staging := "${THISDIR}/lttng-tools:"
 
 SRC_URI:append:feature-sokol-flex-staging = " file://0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch"
 
+STAGING_ENABLED = ""
+STAGING_ENABLED:feature-sokol-flex-staging = "1"
+
 python () {
-    if not d.getVar('MULTILIBS'):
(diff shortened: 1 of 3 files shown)


meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_%.bbappend
```diff
@@ -1,3 +1,11 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/lttng-tools:"
```
abelal, line 5: should we be using the staging featureoverride here? and in the other places?

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_2.13.11.bb
abelal, line 1: Is this an upgrade? The commit doesn't talk about this

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_%.bbappend
kergoth: Yes, thanks!

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_2.13.11.bb
kergoth: Unintentional inclusion, thanks for spotting. I upgraded both components in testing, but weren't the critical elements. Will fix!

abelal (approved):

PR #150, opened 2023-10-05
setup-flex-builddir: exclude flex-vendor-integration for qemu

tarmeh2r:
This resolves the following warning for qemu BSPs: "No bb files in default matched BBFILE_PATTERN_flex-vendor-integration"

JIRA: SB-22633


scripts/setup-flex-builddir
```diff
@@ -261,6 +261,7 @@ setup_builddir () {
     case ${MACHINE} in
         qemu*)
             OPTIONALLAYERS="$OPTIONALLAYERS flex-bsp-$MACHINE"
+            EXCLUDEDLAYERS="flex-vendor-integration"
```
Noor-Ahsan, line 264: this will exclude if we will be including the layer in layer.conf. I think commit message needs to be updated. it remove the layer from setup

abelal:
@tarmeh2r why is this needed even after removing the vendor-integration layer dependency from bsp-common?

tarmeh2r:
@abelal because meta-sokol-flex recommends flex-vendor-integration and it gets included in the build if available. Therefore, we explicitly remove it

scripts/setup-flex-builddir
tarmeh2r: @Noor-Ahsan this will exclude the vendor-integration layer when we use the setup-env script to create the build configuration. Let me add further description to the commit message.

abelal: @tarmeh2r nitpick here. Please don't override excluded layers rather append to it. i.e.
```
EXCLUDEDLAYERS="$EXCLUDEDLAYERS flex-vendor-integration"
```

tarmeh2r: Thanks @abelal I have made the requested changes. @Noor-Ahsan I have also added further explanation in the commit message.

abelal (approved):

tarmeh2r:
We will be creating a separate manifest for qemu that will not pull flex-vendor-integration layer. Therefore, we don't need to exclude it from the scripts explicitly.

kergoth:
I still think there's no reason we can't just enable the argument to suppress any warnings due to no files in flex vendor integration. It's a single variable. Only downside if we won't see warnings if we have no files there unintentionally, but that's a pretty unlikely case which can be easily verified in other ways.

kergoth:
`BBFILE_PATTERN_IGNORE_EMPTY_flex-vendor-integration = "1"`

PR #151, opened 2023-10-17
lttng-tools: fix non-multilib failure

kergoth:
As was diagnosed by @aarslan7  (great work, Arslan!), the configure script was
unconditionally defining the variables for consumerd, but this resulted in a
value of the empty string, not NULL, and the code only checked for non-NULL.

Change the code to actually check if the binary exists with stat before running
it. This is consistent with other uses of these variables in the code, ensures
it isn't used as the empty string, and ensures that it doesn't try to run a
missing binary in general.

JIRA: SB-22858

### DIFF
--- a/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
+++ b/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
@@ -1,19 +1,19 @@
-From 0d09ed08a84ec77e01bede24847f4ced4e99057c Mon Sep 17 00:00:00 2001
+From 30e90ed6f5a0b187d75c0c12304198b01af7e3ea Mon Sep 17 00:00:00 2001
 From: Christopher Larson <chris_larson@mentor.com>
 Date: Fri, 29 Sep 2023 17:07:30 -0500
 Subject: [PATCH] Ensure that the consumerd configure arguments are used
 
-Currently, the --with-consumerd32-bin, --with-consumerd32-libdir, etc arguments
-are not being used by lttng-sessiond, which causes problems with providing a
-well configured multilib tracing environment. Ensure that they are used as the
-default values in sessiond.
+Currently, the --with-consumerd32-bin, --with-consumerd32-libdir, etc
+arguments are not being used by lttng-sessiond, which causes problems
+with providing a well configured multilib tracing environment. Ensure
+that they are used as the default values in sessiond.
 
 Signed-off-by: Christopher Larson <chris_larson@mentor.com>
+
 ---
-Upstream-Status: Pending
 
 diff --git a/src/bin/lttng-sessiond/sessiond-config.c b/src/bin/lttng-sessiond/sessiond-config.c
-index 05e0ef1..d98876d 100644
+index 05e0ef1..e039287 100644
 --- a/src/bin/lttng-sessiond/sessiond-config.c
 +++ b/src/bin/lttng-sessiond/sessiond-config.c
 @@ -51,14 +51,14 @@ struct sessiond_config sessiond_config_build_defaults = {
@@ -22,16 +22,16 @@ index 05e0ef1..d98876d 100644
  	.consumerd32_path.value =		NULL,
 -	.consumerd32_bin_path.value =		NULL,
 -	.consumerd32_lib_dir.value =		NULL,
-+	.consumerd32_bin_path.value =		DEFAULT_USTCONSUMERD32_BIN,
-+	.consumerd32_lib_dir.value =		DEFAULT_USTCONSUMERD32_LIBDIR,
++	.consumerd32_bin_path.value =		DEFAULT_USTCONSUMERD32_BIN[0] != '\0' ? DEFAULT_USTCONSUMERD32_BIN : NULL,
++	.consumerd32_lib_dir.value =		DEFAULT_USTCONSUMERD32_LIBDIR[0] != '\0' ? DEFAULT_USTCONSUMERD32_LIBDIR : NULL,
  	.consumerd32_err_unix_sock_path.value = NULL,
  	.consumerd32_cmd_unix_sock_path.value = NULL,
  
  	.consumerd64_path.value =		NULL,
 -	.consumerd64_bin_path.value =		NULL,
 -	.consumerd64_lib_dir.value =		NULL,
-+	.consumerd64_bin_path.value =		DEFAULT_USTCONSUMERD64_BIN,
-+	.consumerd64_lib_dir.value =		DEFAULT_USTCONSUMERD64_LIBDIR,
++	.consumerd64_bin_path.value =		DEFAULT_USTCONSUMERD64_BIN[0] != '\0' ? DEFAULT_USTCONSUMERD64_BIN : NULL,
++	.consumerd64_lib_dir.value =		DEFAULT_USTCONSUMERD64_LIBDIR[0] != '\0' ? DEFAULT_USTCONSUMERD64_LIBDIR : NULL,
  	.consumerd64_err_unix_sock_path.value = NULL,
  	.consumerd64_cmd_unix_sock_path.value = NULL,
  
--- a/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_%.bbappend
+++ b/meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools_%.bbappend
@@ -4,7 +4,9 @@
 
 FILESEXTRAPATHS:prepend:feature-sokol-flex-staging := "${THISDIR}/lttng-tools:"
 
-SRC_URI:append:feature-sokol-flex-staging = " file://0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch"
+SRC_URI:append:feature-sokol-flex-staging = "\
+    file://0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch \
+"
 
 STAGING_ENABLED = ""
 STAGING_ENABLED:feature-sokol-flex-staging = "1"


aarslan7 (approved):

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
```diff
@@ -1,17 +1,50 @@
...
+ 			if (the_config.consumerd64_bin_path.value &&
++					stat(the_config.consumerd64_bin_path.value, &st) == 0 &&
+ 					the_ustconsumer64_data.pid == 0 &&
+ 					cmd_ctx->lsm.cmd_type != LTTNG_REGISTER_CONSUMER) {
+ 				pthread_mutex_unlock(&the_ustconsumer64_data.pid_mutex);
+@@ -1314,6 +1316,7 @@ static int process_client_msg(struct command_ctx *cmd_ctx, int *sock,
+ 			/* 32-bit */
+ 			pthread_mutex_lock(&the_ustconsumer32_data.pid_mutex);
+ 			if (the_config.consumerd32_bin_path.value &&
++					stat(the_config.consumerd32_bin_path.value, &st) == 0 &&
```
aarslan7, line 44: when the code reaches this point, `the_config.consumerd32_bin_path.value` has been resolved to `/home/root/` so we still get the issue as this path exists. so maybe add a check here that it is a regular file, and not a directory?

aarslan7 (changes requested):

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
kergoth: Good catch, thanks

kergoth: Hmm, when I build, I'm not seeing the values set to /home/root/, but the empty string. Are you able to reproduce this aspect?

aarslan7: empty paths are resolved to `/home/root` at runtime when config is parsed. so that will look fine at build time, but at runtime, it will still throw the error

aarslan7: https://github.com/lttng/lttng-tools/blob/v2.13.9/src/bin/lttng-sessiond/sessiond-config.c#L475

umairsair:
just wondering why we don't take the easier path.. initialize something like below.. same for other locations..

```
.consumerd32_bin_path.value =		(strlen(DEFAULT_USTCONSUMERD32_BIN) ? DEFAULT_USTCONSUMERD32_BIN : NULL),
```

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
```diff
@@ -1,17 +1,53 @@
...
++	struct stat st;
+ 	bool need_tracing_session = true;
+ 	bool need_domain;
+ 	bool need_consumerd;
+@@ -1285,7 +1286,8 @@ static int process_client_msg(struct command_ctx *cmd_ctx, int *sock,
+ 			/* 64-bit */
+ 			pthread_mutex_lock(&the_ustconsumer64_data.pid_mutex);
+ 			if (the_config.consumerd64_bin_path.value &&
+-					the_ustconsumer64_data.pid == 0 &&
++					stat(the_config.consumerd64_bin_path.value, &st) == 0 &&
```
umairsair, line 36: if user provided wrong path as argument to sessiond, we'll be ignoring it instead of properly erroring out?

kergoth:
> just wondering why we don't take the easier path.. initialize something like below.. same for other locations..
> 
> ▾
> ```
> .consumerd32_bin_path.value =		(strlen(DEFAULT_USTCONSUMERD32_BIN) ? DEFAULT_USTCONSUMERD32_BIN : NULL),
> ```

Good idea! I do think it's still worth having the other check, is it's consistent with the way they do things elsewhere, but this avoids the issue entirely.

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
kergoth: Good point! Both kernel consumerd and sessiond paths show errors if it fails to find a valid path or has a path it can't access, whereas the 32 and 64 bit consumerd will end up getting an execl() failure with a less pleasant error message for the user. This change makes it silent rather than trying and failing to spawn, which is worse. I'll drop this, since there's no further error checking on these values done elsewhere, better to let the spawn fail than to show the user no feedback. Ideally it'd probably check for invalid or missing paths earlier on in the config handling.

kergoth: Good point! Both kernel consumerd and sessiond paths show errors if it fails to find a valid path or has a path it can't access, whereas the 32 and 64 bit consumerd will end up getting an execl() failure with a less pleasant error message for the user. This change makes it silent rather than trying and failing to spawn, which is worse. I'll drop this, since there's no further error checking on these values done elsewhere, better to let the spawn fail than to show the user no feedback. Ideally it'd probably check for invalid or missing paths earlier on in the config handling.

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0002-Don-t-resolve-empty-strings-in-resolve_path.patch
```diff
@@ -0,0 +1,26 @@
+From 60bad37b01dfa851bc4b4a7fbb9a7086c7453251 Mon Sep 17 00:00:00 2001
+From: Christopher Larson <chris_larson@mentor.com>
+Date: Thu, 19 Oct 2023 11:08:18 -0500
+Subject: [PATCH] Don't resolve empty strings in resolve_path
+
+If a resolved path is set to the empty string, this would resolve it to
+$HOME. Instead, leave it as is so the caller can deal with it.
```
umairsair, line 7: it resolves to CWD, not $HOME.. we observed it to be $HOME because CWD was $HOME..

Also I am not really sure if we need to fix it.. which known problematic scenario are we fixing? if its really an issue, then IMHO it should go upstream.. just random thoughts so you can simply ignore it..

umairsair (approved):

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0002-Don-t-resolve-empty-strings-in-resolve_path.patch
kergoth: Fair point, I'll look into submitting a few things in addition to this upstream, and drop the second patch in this PR, as I do think they should really have a cleaner error message if given incorrect consumerd paths.

kergoth: Fair point, I'll look into submitting a few things in addition to this upstream, and drop the second patch in this PR, as I do think they should really have a cleaner error message if given incorrect consumerd paths.

aarslan7 (approved):

aarslan7:
the path is now showing as `Unknown`. I guess this is fine. Enable channel doesn't get stuck:
```
root@imx8qxpmek-flex:~# lttng-sessiond -vvv --verbose-consumer 2>&1 | tee log
DBG1 - 06:34:58.259327722 [Main]: [sessiond configuration]
DBG1 - 06:34:58.259473976 [Main]:       version                        2.13.9
DBG1 - 06:34:58.259493477 [Main]:       verbose:                       3
DBG1 - 06:34:58.259511352 [Main]:       verbose consumer:              1
DBG1 - 06:34:58.259528103 [Main]:       quiet mode:                    False
DBG1 - 06:34:58.259547853 [Main]:       agent_tcp_port:                [5345, 5354]
DBG1 - 06:34:58.259568354 [Main]:       application socket timeout:    5
DBG1 - 06:34:58.259584855 [Main]:       no-kernel:                     False
DBG1 - 06:34:58.259600980 [Main]:       background:                    False
DBG1 - 06:34:58.259617106 [Main]:       daemonize:                     False
DBG1 - 06:34:58.259633481 [Main]:       signal parent on start:        False
DBG1 - 06:34:58.259649357 [Main]:       tracing group name:            tracing
DBG1 - 06:34:58.259665482 [Main]:       kmod_probe_list:               None
DBG1 - 06:34:58.259681857 [Main]:       kmod_extra_probe_list:         None
DBG1 - 06:34:58.259698108 [Main]:       rundir:                        /var/run/lttng
DBG1 - 06:34:58.259714233 [Main]:       application socket path:       /var/run/lttng/lttng-ust-sock-8
DBG1 - 06:34:58.259730859 [Main]:       client socket path:            /var/run/lttng/client-lttng-sessiond
DBG1 - 06:34:58.259746984 [Main]:       wait shm path:                 /lttng-ust-wait-8
DBG1 - 06:34:58.259763235 [Main]:       health socket path:            /var/run/lttng/sessiond-health
DBG1 - 06:34:58.259779360 [Main]:       LTTNG_UST_CLOCK_PLUGIN:        None
DBG1 - 06:34:58.259795236 [Main]:       pid file path:                 /var/run/lttng/lttng-sessiond.pid
DBG1 - 06:34:58.259814611 [Main]:       lock file path:                /var/run/lttng/lttng-sessiond.lck
DBG1 - 06:34:58.259831987 [Main]:       session load path:             None
DBG1 - 06:34:58.259854238 [Main]:       agent port file path:          /var/run/lttng/agent.port
DBG1 - 06:34:58.259875113 [Main]:       consumerd32 path:              /var/run/lttng/ustconsumerd32
DBG1 - 06:34:58.259895114 [Main]:       consumerd32 bin path:          Unknown
DBG1 - 06:34:58.259915364 [Main]:       consumerd32 lib dir:           Unknown
```
Enable channel doesn't get stuck:
```
root@imx8qxpmek-flex:~# lttng create a
Spawning a session daemon
[   24.741874] lttng_kretprobes: loading out-of-tree module taints kernel.
[   24.837476] LTTng: Loaded modules v2.13.9 (Nordicité)
Session a created.
Traces will be output to /home/root/lttng-traces/a-20231020-063155
root@imx8qxpmek-flex:~# lttng enable-channel diag_channel_userspace -u --buffers-pid -s a
ust channel diag_channel_userspace enabled for session a
```

umairsair (approved):

PR #152, opened 2023-11-07
Merge kirkstone into develop

tarmeh2r:


tarmeh2r:
@kergoth given that develop is compatible with mickledore, is merging kirkstone into develop the way to go? Or is there a better strategy that we can adopt here? Please let us know if you have any reservations on this PR.

kergoth:
Yes, we should merge kickstone and then make any changes necessary to adjust those for later releases.

kergoth (approved):
LGTM, didn't review each change one by one since we did that when merging them to kirkstone in the first place.

PR #156, opened 2024-01-30
README.md: Reword README for accuracy

tammerOfTheGods:
Correct reference to three sub-layers when there are four.


tammerOfTheGods:
@Noor-Ahsan - please review

Noor-Ahsan:
@tammerOfTheGods kindly correct the branch. I think we are using Kirkstone for this layer for hycon

tammerOfTheGods:
> @tammerOfTheGods kindly correct the branch. I think we are using Kirkstone for this layer for hycon

Will do.  I have found an additional reference to meta-flex-private that should be removed as well.

PR #157, opened 2024-01-31
conf/multiconfig: add support for multiconfig

ansar-rasool:
This updates setup-flex-builddir script to by default create multiconfigs for dom0 and domU. This also adds sample conf files for both multiconfigs.

JIRA: https://jira.alm.mentorg.com/browse/AT1-8

### DIFF
--- a/scripts/setup-flex-builddir
+++ b/scripts/setup-flex-builddir
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-2.0
 # ---------------------------------------------------------------------------------------------------------------------
 #
-# Copyright 2022 Siemens
+# Copyright 2024 Siemens
 #
 # This file is licensed under the terms of the GNU General Public License
 # version 2.  This program  is licensed "as is" without any warranty of any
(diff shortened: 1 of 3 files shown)


scripts/setup-flex-builddir
```diff
@@ -275,7 +275,7 @@ setup_builddir () {
         existing_builddir=0
     fi
 
-    mkdir -p $BUILDDIR/conf
+    mkdir -p $BUILDDIR/conf/multiconfig
```
Noor-Ahsan, line 278: I think we don't need multiconfig in the build folder. If we have it in layer as well it will work. Can you check that. If it is true then we don't need this change and other changes in this file

ansar-rasool: Yes, that's true. we can have multiconfig in our layer as well. But we are separating out build configurable parameters for dom0 and domu in their respective conf files instead of having them all mixed up in local.conf file. That's why we put these multiconfig files in build folder. I think having user to change build parameters from layer dir does not looks feasible. Thoughts?

ansar-rasool:
Closing this PR as we have moved the multiconfig support to meta-flex-virtualization layer. We will continue to use main branch of meta-sokol-flex repo.

PR #158, opened 2024-02-01
README.md: Reword README for accuracy

tammerOfTheGods:
Correct reference to three sub-layers when there are four.

Signed-off-by: Thomas Brown <ThomasBrown@Siemens.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # OpenEmbedded/Yocto layers for Sokol Flex OS
 
-This repository holds the layers needed by our Yocto builds of Sokol Flex OS. At
-this time there are three such layers:
+This repository holds the layers needed by our Yocto builds of Sokol Flex OS.
+Currently the Sokol Flex OS layers are:
 
 - meta-sokol-flex-distro: The Sokol Flex OS distribution layer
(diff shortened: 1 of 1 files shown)


tammerOfTheGods:
@Noor-Ahsan - please review.  I had to create a new fork that included the kirkstone branch.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -174,9 +174,6 @@ INHERIT += "deploy-license-manifest"
 # current configuration
 INHERIT += "flex-updates-check"
 
-# Check PDK license
```
Noor-Ahsan: what will happen if we keep this and only remove the layer?

Noor-Ahsan: I am asking because this is shared layer so this will be used for the BSP where we need to enable license

tammerOfTheGods: No problems were observed with the INHERIT in place.  I'll drop the change.

PR #159, opened 2024-03-15
setup-rh: Make combined script for both RHEL and Rocky host and update accordingly

jainash:
Made combined script for both RHEL and Rocky host and updated accordingly.

Enabling required repo's to get install rpcgen package.

And due to lower make and python version we are unable to make build succesful so installing required version and creating symbolic link.

JIRA ID: SB-23264


Noor-Ahsan:
@jainash just add error w.r.t make and python which you were getting in commit summary

PR #161, opened 2024-04-01
local.conf.sample: elaborate TCMODE:sokol-flex comment

hashraf91:
JIRA: SB-23194

### DIFF
--- a/conf/local.conf.sample
+++ b/conf/local.conf.sample
@@ -12,7 +12,8 @@ MACHINE ??= "qemux86"
 # where FLEXDIR is the path to your Sokol Flex OS workspace
 #TOOLCHAINS_PATH ?= "${FLEXDIR}/toolchains"
 
-# Use the oe/yocto-built toolchain rather than the external
+# Uncomment to use the oe/yocto-built toolchain rather than the external
+# Sourcery CodeBench toolchain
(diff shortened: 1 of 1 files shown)


conf/local.conf.sample
```diff
@@ -12,7 +12,8 @@ MACHINE ??= "qemux86"
 # where FLEXDIR is the path to your Sokol Flex OS workspace
 #TOOLCHAINS_PATH ?= "${FLEXDIR}/toolchains"
 
-# Use the oe/yocto-built toolchain rather than the external
+# Uncomment to use the oe/yocto-built toolchain rather than the external (e.g.
+# CodeBench toolchain)
```
sumaiyyafarooq: @Noor-Ahsan  Should it be CodeBench or Sourcery?

Noor-Ahsan: @sumaiyyafarooq I checked the reference manual and it does not talk about sourcery toolchain. we have following
meta-sourcery Contains files that support the Sourcery CodeBench
toolchain.

So lets use "external Sourcery CodeBench toolchain". Thoughts?

sumaiyyafarooq: It is fine.

hashraf91: Updated.

PR #162, opened 2024-04-09
sokol-flex: allow version configuration from bsp layer

hashraf91:
Hycon BSP will be following new version scheme.

JIRA: SB-23304


Noor-Ahsan:
can't we include this in machine.conf? I know you have added it in distro folder but lets add it in machine file if there is no technical issue

hashraf91:
> can't we include this in machine.conf? I know you have added it in distro folder but lets add it in machine file if there is no technical issue

The soft variable assignments can be updated from machine.conf but not the hard variable assignments. Most of the assignments are hard assignments in the sokol-flex.conf. So, I think we'll need this.

hashraf91:
Closing this PR, the change is made through this PR: https://github.com/MentorEmbedded/meta-flex-ref-bsps/pull/214

PR #165, opened 2024-04-24
virtualization-layer: provide option to add extra-space

hashraf91:
The user may need extra space when virtualization feature is enabled. Even for our QVTs where we pull ubuntu docker and try installing some packages in it, we need around 200M of additional space.

JIRA-ID: SB-23366

### DIFF
--- a/meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
+++ b/meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
@@ -1,2 +1,5 @@
 # Comment out the following to disable virtualization support
 USER_FEATURES += "virtualization"
+
+# Uncomment to add extra-space to rootfs (default: 204800KB) for virtualization only
+#IMAGE_ROOTFS_EXTRA_SPACE:append = "${@bb.utils.contains('USER_FEATURES', 'virtualization', ' + 204800', '', d)}"


meta-sokol-flex-distro/dynamic-layers/virtualization-layer/conf/local.conf.append
```diff
@@ -1,2 +1,5 @@
 # Comment out the following to disable virtualization support
 USER_FEATURES += "virtualization"
+
+# Uncomment to add extra-space to rootfs (default: 204800KB) for virtualization only
+#IMAGE_ROOTFS_EXTRA_SPACE:append = "${@bb.utils.contains('USER_FEATURES', 'virtualization', ' + 204800', '', d)}"
```
ahsanhussain, line 5: This should not be added to the distro layer IMO, as we have resizer available for physical media. This definitely shouldn't be made specific to virtualization because extra space could be needed for any purpose. I think you can do this generically under the Qemu BSPs' local.conf.append.

I'm not sure if there's a way to increase the image's block storage and FS size under Qemu. If there's a way to do this live, that would be better.

hashraf91: > I think you can do this generically under the Qemu BSPs' local.conf.append.

local.conf.append.qemu is not being included for Hycon BSP. I can include local.conf.append.qemu for Hycon as well but this will bring the following in local.conf as well:
```
# By default qemu will build with a builtin VNC server where graphical output can be
# seen. The two lines below enable the SDL backend too. By default libsdl2-native will
# be built, if you want to use your host's libSDL instead of the minimal libsdl built
# by libsdl2-native then uncomment the ASSUME_PROVIDED line below.
PACKAGECONFIG:append:pn-qemu-system-native = " sdl"
PACKAGECONFIG:append:pn-nativesdk-qemu = " sdl"
#ASSUME_PROVIDED += "libsdl2-native"
```
Alternatively, we can include this change only for Hycon BSP, suggestions?

> I'm not sure if there's a way to increase the image's block storage and FS size under Qemu. If there's a way to do this live, that would be better.

If by "live", you mean either on running Flex-OS on QEMU or in QEMU parameters, that won't be possible. Although, we can resize the ext4 development-image before copying it into Hycon workspace from command line (resize2fs).

ahsanhussain: Had a call with Haseeb on this. To get the release out we're currently setting this in vendor conf which updates the distro config for the machine.

hashraf91: Closing this PR in favor of: https://github.com/MentorEmbedded/meta-flex-ref-bsps/pull/219.

PR #166, opened 2024-05-03
kernel-fitimage.bbclass: Remove from staging area

ahsanhussain:
The current core layer class has no diff from the one being removed

JIRA-ID: MCF-27


aarslan7 (approved):
LGTM

PR #168, opened 2024-05-03
capdebuginfo: Remove it as not required

jainash:
This needs to be removed as no one is using it.

JIRA ID: SB-23398


Noor-Ahsan:
Let's discuss this with Awais and make a decesion. As remove from here will remove it from other BSPs as well and we have to update the docs

jainash:
we will not remove it from meta data so closing this PR as it is.

PR #169, opened 2024-06-03
librsvg-2.52.10: use OSS compliant source

hashraf91:
We maintain modified sources in repo:
https://github.com/MentorEmbedded/flex_package_source/

These sources are OSS compliant and remove objectionable binaries as compared to the original sources.

JIRA-ID: SB-23487


Noor-Ahsan:
You can remove openembedded-layer: from commit summary

PR #172, opened 2024-08-27
meta-sokol-flex: preliminary changes for Yocto 5 support

ahsanhussain:


meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
```diff
@@ -8,6 +8,8 @@ arguments are not being used by lttng-sessiond, which causes problems
 with providing a well configured multilib tracing environment. Ensure
 that they are used as the default values in sessiond.
 
+Upstream-Status: Inappropriate [OE specific]
```
kergoth, line 11: I don't believe this is the correct status. This should probably be upstreamed, or at least discussed with upstream. Both lttng proper and oe-core.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
```diff
@@ -2,19 +2,20 @@
...
 SUPPORT_URL = "https://support.sw.siemens.com/"
 BUG_REPORT_URL = "https://support.sw.siemens.com/"
 
 # Distro and release versioning
-DISTRO_VERSION = "13"
+YEAR_VERSION = "2024"
+QUARTER_VERSION = "04"
+UPDATE_VERSION = "0"
+RELEASE_VERSION = "${YEAR_VERSION}.${QUARTER_VERSION}.${UPDATE_VERSION}"
+DISTRO_VERSION = "${RELEASE_VERSION}"
```
hashraf91, line 18: Is there any plan to change DISTRO from sokol-flex to flex-os in this release?
@Noor-Ahsan 

This will be required to remove all of the re-branding changes which were done in BSP layer:
https://github.com/MentorEmbedded/meta-flex-ref-bsps/tree/develop/hycon/meta-flex-bsp-hycon-arm64/recipes-core and in machine.conf

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
ahsanhussain: @kergoth since you're the author of this patch, can you initiate the discussion. I'll update the status.

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
Noor-Ahsan: Awais was asking for an effort for updating the distro. I think its couple of days effort. Can you guys confirm this?

We have upstream layer name as meta-sokol-flex as well. We need to check if he wants to change that as well. keep it mind its upstream layer.

hashraf91: Yes, I think couple of days would be required.
The major change-set will be to rename all the usages of DISTROOVERRIDE :sokol-flex in different layers.

Moreover, re-name of following layers and their overrides will also be required:
sokol-flex-distro, sokol-flex-support, sokol-flex-staging and sokol-flex-common.

meta-sokol-flex-staging/recipes-kernel/lttng/lttng-tools/0001-Ensure-that-the-consumerd-configure-arguments-are-us.patch
Noor-Ahsan: I am merging it. we will handle it in cleanup task

meta-sokol-flex-distro/conf/distro/sokol-flex.conf
ahsanhussain: Think we can get away without changing the layer names but of-course something to double check with PO.

PR #175, opened 2024-09-16
Virtualization and EWAOL integration changes

ahsanhussain:


Noor-Ahsan:
Looks good

meta-flex-distro/conf/distro/flex-os.conf
```diff
@@ -348,6 +349,11 @@ PUBLIC_REPOS ?= "meta-sokol-flex ${FORKED_REPOS}"
 
 # Define a location for placing external artifacts to be used by the build
 FLEX_EXTERNAL_ARTIFACTS ?= "${TOPDIR}/flex-external-artifacts"
+
+# Add kernel module development capability to SDK
+TOOLCHAIN_HOST_TASK_EXTRA:append = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-debugging', ' nativesdk-flex nativesdk-bison', '', d)}"
+TOOLCHAIN_TARGET_TASK_EXTRA:append = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-debugging', ' flex-kernel-devsrc', '', d)}"
```
hashraf91: Should the `flex-kernel-devsrc` recipe be moved to distro layer as well?

meta-flex-distro/conf/distro/flex-os.conf
```diff
@@ -348,6 +349,11 @@ PUBLIC_REPOS ?= "meta-sokol-flex ${FORKED_REPOS}"
...
+SDK_POST_INSTALL_COMMAND = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-debugging', '. $target_sdk_dir/environment-setup-armv8a-oe-linux && cd $target_sdk_dir/sysroots/armv8a-oe-linux/usr/src/kernel && make scripts prepare modules_prepare', '', d)}"
```
hashraf91: Can we make `armv8a-oe-linux` generic here? Maybe by using ${REAL_MULTMACH_TARGET_SYS}.

meta-flex-distro/conf/distro/flex-os.conf
```diff
@@ -373,6 +379,9 @@ include conf/distro/include/flex-ml.conf
 # BSP related distro changes
 include conf/distro/include/flex-distro.conf
 
+# EWAOL integration
+require ${@bb.utils.contains('USER_FEATURES', 'flex-ewaol-baremetal', 'conf/distro/include/ewaol-flex-integration.conf', '', d)}
```
hashraf91, line 379: We use DISTRO_FEATURES usually instead of USER_FEATURES in such conditions. Functionally it's fine, just for consistency.

meta-flex-distro/conf/distro/flex-os.conf
ahsanhussain: Whoops! This shouldn't have made it to the distro or the common BSP layers. I'll update in a separate commit :+1:

meta-flex-distro/conf/distro/flex-os.conf
ahsanhussain: This is distro conf. DISTRO_FEATURES are still being parsed, and feature_overrides.bbclass works on event bb.event.ConfigParsed so this user feature wouldn't have made it to DISTRO_FEATURES yet. There are other such cases too e.g. https://github.com/MentorEmbedded/meta-flex-security/blob/develop/meta-flex-selinux/conf/distro/include/flex-security-selinux.conf#L10

meta-flex-distro/conf/distro/flex-os.conf
ahsanhussain: I don't have a strong opinion here. To me kernel recipes usually reside in the BSP layers, in this case the common BSP layer which seems fine to me

Noor-Ahsan: I think what he wanna say that as this is upstream layer so if someone will use this layer with his distro and enable flex-debugging he will get an error for missing flex-kernel-devsrc recipe.

hashraf91: Yes, exactly. This makes distro layer dependent on common BSP layer.

In our case, common BSP layer depends on distro layer and we use them together. So, we don't see any issue but this requires that distro and common BSP layers are actually interdependent. IMO, we should keep the dependency unidirectional i.e. common BSP layer depends on distro layer only, not the other way around which would require that we move this recipe to distro layer. So that distro layer is complete in itself.

meta-flex-distro/conf/distro/flex-os.conf
hashraf91: Interesting. Perhaps worth mentioning somewhere in comments or commit message.

meta-flex-distro/conf/distro/flex-os.conf
```diff
@@ -348,6 +349,11 @@ PUBLIC_REPOS ?= "meta-sokol-flex ${FORKED_REPOS}"
 
 # Define a location for placing external artifacts to be used by the build
 FLEX_EXTERNAL_ARTIFACTS ?= "${TOPDIR}/flex-external-artifacts"
+
+# Add kernel module development capability to SDK
+TOOLCHAIN_HOST_TASK_EXTRA:append = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-debugging', ' nativesdk-flex nativesdk-bison', '', d)}"
+TOOLCHAIN_TARGET_TASK_EXTRA:append = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-debugging', ' flex-kernel-devsrc', '', d)}"
+SDK_POST_INSTALL_COMMAND = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-debugging', '. $target_sdk_dir/environment-setup-${TUNE_PKGARCH}${TARGET_VENDOR}-${TARGET_OS} && cd $target_sdk_dir/sysroots/armv8a-oe-linux/usr/src/kernel && make scripts prepare modules_prepare', '', d)}"
```
hashraf91: This is only partially updated. The second command `cd $target_sdk_dir/sysroots/armv8a-oe-linux/usr/src/kernel` should be updated as well.

hashraf91 (changes requested):

hashraf91 (approved):

meta-flex-distro/conf/distro/flex-os.conf
ahsanhussain: Added a comment

PR #181, opened 2024-10-04
DRAFT: qemu-system-native: fix mouse co-ordinates scaling issue

hashraf91:
This is in continuation of commit 59cc33db.

The previous commit solved the issue with SDL,GL but since scaling issue was only related to opengl, the scaling shouldn't have been applied otherwise. So, conditionally scale the mouse co-ordinates only for SDL,GL.

Moreover, moved the logic to common place and now the scaling isn't actually required, if we just choose the correct surface dimension based on opengl context.

JIRA-ID: SB-23711


hashraf91:
@kergoth I checked with qemu-9.0.2 in oe-core:master-next and this issue is still there. Although, I have yet to check in QEMU master.

@Noor-Ahsan we'll have to submit this updated patch to qemu-s2s repository as well.

PR #187, opened 2024-10-21
RFS Optimization

ahsanhussain:
We use a separate kernel image for our BSPs so there's no need to install an image in /boot. Doing this change for development-image as that's the only image we use with Flex OS. This saves around 62M on ANA BSP from the RFS size

JIRA-ID: AT1-1825


Noor-Ahsan (approved):

meta-flex-support/recipes-core/images/development-image.bb
```diff
@@ -5,22 +5,22 @@
...
 
-IMAGE_FEATURES = "${EXTRA_IMAGE_FEATURES}"
-EXTRA_IMAGE_FEATURES = "debug-tweaks codebench-debug ssh-server-openssh"
-IMAGE_FEATURES:append:feature-tracing = " tools-profile"
+IMAGE_FEATURES = " \
+    codebench-debug \
+    debug-tweaks \
+    package-management \
+    ssh-server-openssh \
+"
```
hashraf91, line 13: Have you check the default value of EXTRA_IMAGE_FEATURES if it is bringing any other feature? In that case, we'll have to empty that variable as well.

hashraf91: I checked the environment, this is not an issue. This PR is good to merge.

PR #189, opened 2024-10-22
development-image: do not overwrite IMAGE_FEATURES default value

hashraf91:
Overwriting the image features default value excludes the EXTRA_IMAGE_FEATURES. So, use append here to add features to the IMAGE_FEATURES.

JIRA-ID: SB-23686


ahsanhussain:
This is not quite the issue here as the same is done in oe-core/meta/conf/bitbake.conf; the issue is hard assignment of IMAGE_FEATURES in the image recipe which overrides all the assignments from lower priority layers. Just make the assignment an append and I believe we'd be good to go.

ahsanhussain (approved):

PR #190, opened 2024-10-29
flex-os.conf: Fix ARCHIVE_RELEASE_VERSION

pragow0k:


hashraf91 (approved):

meta-flex-distro/conf/distro/flex-os.conf
```diff
@@ -17,7 +17,7 @@ QUARTER_VERSION = "04"
 UPDATE_VERSION = "0"
 RELEASE_VERSION = "${YEAR_VERSION}.${QUARTER_VERSION}.${UPDATE_VERSION}"
 DISTRO_VERSION = "${RELEASE_VERSION}"
-ARCHIVE_RELEASE_VERSION = "${DISTRO_VERSION}.${BSP_VERSION}.${PATCH_VERSION}"
+ARCHIVE_RELEASE_VERSION = "${DISTRO_VERSION}"
```
hashraf91, line 20: I think we can also remove these variables now:
BSP_VERSION
PATCH_VERSION

pragow0k: removed it from the script

hashraf91 (dismissed):

PR #191, opened 2024-11-06
drop the TEMPLATECONF and bmaptool from installer and fix SCRIPTS_VERSION and QUARTER_VERSION

hashraf91:
* archive-release: drop the PDK_* variables as licensing not required
* archive-release: standardize template directory for TEMPLATECONF
* archive-release: drop TEMPLATECONF packaging for installer
* archive-release: drop bmaptool packaging for installer
* flex-os.conf: mark backwards incompatibility for new release
* flex-os.conf: make QUARTER_VERSION a single digit only
* scripts/setup-rh: remove -x flag from production

JIRA-ID: SB-23870


ahsanhussain:
Sorry haven't taken a look myself, do you know why does archive-release recipe need TEMPLATECONF? Also, should we use a common value for TEMPLATECONF across scripts/setup-flex-builddir and this? It's being populated at both places

hashraf91:
> Sorry haven't taken a look myself, do you know why does archive-release recipe need TEMPLATECONF? Also, should we use a common value for TEMPLATECONF across scripts/setup-flex-builddir and this? It's being populated at both places

I do not know why TEMPLATECONF was introduced separately for the installer but I can guess from the changes that archive-release.bb is doing. It is editing bblayers.conf.sample and appending all the layers of build in it. It is editing local.conf.sample and adjusting MACHINE/DISTRO variables and adding PDK_DISTRO_VERSION variable.

I tried by removing the TEMPLATECONF of the installer, so that it uses the same TEMPLATECONF that is available in meta-flex (meta-flex/meta-flex-distro/conf/templates/default). The bblayers.conf can still resolve dependencies and contain all layers. MACHINE and DISTRO variables are also adjusted in local.conf. The only difference observed was in local.conf that PDK_DISTRO_VERSION was not defined.

So, it seems like this separate TEMPLATECONF is being used to add installer specific variables to the local.conf only. Is this worth it? Should I remove the TEMPLATECONF packaging into installer and define PDK_DISTRO_VERSION somewhere else?

Noor-Ahsan:
As far as PDK_DISTRO_VERSION is concern I think it was introduced in this commit https://github.com/MentorEmbedded/mel-meta-mentor-private/commit/9cf31b4ddd5d4f401bc98c9d20521efa8483554c#diff-f331609cebc0e03573068ae337eb16572bec5c0c1baeb2314259f38775422c03R22. it was for licensing. as we have removed licensing, I think we can remove it from archive release as well. Let me further look into about TEMPLATECONF.

Noor-Ahsan:
Here is the logic for templates https://github.com/MentorEmbedded/meta-sokol-flex/commit/e94457354162424d2e4dca433e58237faa02f00a. I think the idea was how we can reproduce the binary image that is part of the installer. For that we saved bbalyers.conf and local.conf with which that initial image was created. this is 12 years back logic :).

hashraf91:
> Here is the logic for templates [e944573](https://github.com/MentorEmbedded/meta-sokol-flex/commit/e94457354162424d2e4dca433e58237faa02f00a). I think the idea was how we can reproduce the binary image that is part of the installer. For that we saved bbalyers.conf and local.conf with which that initial image was created. this is 12 years back logic :).

@noor-ahsan should we keep the template as part of installer if we need to identify the machine configuration with which the binaries are created?

Noor-Ahsan:
I think there is no need to reproduce the images provided by the installer. So we can drop this template. we should test it thoroughly though

hashraf91:
Generated the archive_images (`bitbake -c archive_images archive-release`), the `ana-arm64-flex-2024.4.0.tar` now contains following artifacts:
```
tar -tf ana-arm64-flex-2024.4.0.tar 
ana-arm64-flex/2024.4.0/binary/development-image.cpio
ana-arm64-flex/2024.4.0/binary/Image
ana-arm64-flex/2024.4.0/binary/package.manifest
ana-arm64-flex/2024.4.0/binary/license.manifest
ana-arm64-flex/2024.4.0/binary/image_license.manifest
ana-arm64-flex/2024.4.0/binary/runqemu
ana-arm64-flex/2024.4.0/binary/development-image.qemuboot.conf
```

pragow0k:
Verified this PR:
(updated scripts/setup-flex-builddir)
(removed conf dir)


```
[rhel8@rhel8 work_hycon]$ source ./meta-flex/setup-environment
You had no local.conf file. This configuration file has therefore been
created for you with some default values.
You had no bblayers.conf file. This configuration file has therefore been
created for you with some default values.
Locating layers and bitbake...done
Parsing layer configuration files...done
Processing layers...done
Checking for base layers...core
Determining layers to include for DISTRO 'flex-os'...core,flex-common,flex-distro,flex-staging,flex-support,meta-python,openembedded-layer
Determining layers to include for MACHINE 'hycon-arm64-flex'...core,flex-bsp-common,flex-bsp-hycon-arm64-flex,flex-common,flex-distro,flex-support
Checking for optional layers...

Common bitbake target is:

    development-image

[rhel8@rhel8 build]$ source setup-environment

Common bitbake target is:

    development-image

[rhel8@rhel8 build]$ ls conf/
bblayers.conf  conf-notes.txt  local.conf  templateconf.cfg
[rhel8@rhel8 build]$ ls ../hycon-arm64-flex/
binary
```

Noor-Ahsan (approved):

PR #194, opened 2024-12-09
flex-os.conf: virtualization: increase extra-space in rootfs

jainash:
300M is not sufficient for container tracing validation so we need
to have atleast 700M in rootfs. So, incresing rootfs to 700M.

Moreover changed the condition to add space to rootfs only when
virtualization feature is enabled in absence of ewaol feature
because CASSINI_ROOTFS_EXTRA_SPACE add enough space in case ewaol
feature is enabled.

JIRA-ID: SB-23878

### DIFF
--- a/meta-flex-distro/conf/distro/flex-os.conf
+++ b/meta-flex-distro/conf/distro/flex-os.conf
@@ -205,8 +205,8 @@ SYSTEMD_DEFAULT_TARGET ?= '${@bb.utils.contains_any("IMAGE_FEATURES", ["x11-base
 # Prefer docker-moby to docker-ce by default
 PREFERRED_PROVIDER_virtual/docker = "docker-moby"
 
-# Add extra-space to rootfs (default: 204800KB) for virtualization
-IMAGE_ROOTFS_EXTRA_SPACE:append = "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', ' + 307200', '', d)}"
+# Add extra-space to rootfs (default: 716800KB) for virtualization in absence of flex-ewaol-baremetal
+IMAGE_ROOTFS_EXTRA_SPACE:append:feature-virtualization = "${@bb.utils.contains('USER_FEATURES', 'flex-ewaol-baremetal', '', ' + 716800', d)}"
 
 # Additional package groups
 #


hashraf91 (approved):

meta-flex-distro/conf/distro/flex-os.conf
```diff
@@ -205,8 +205,8 @@ SYSTEMD_DEFAULT_TARGET ?= '${@bb.utils.contains_any("IMAGE_FEATURES", ["x11-base
 # Prefer docker-moby to docker-ce by default
 PREFERRED_PROVIDER_virtual/docker = "docker-moby"
 
-# Add extra-space to rootfs (default: 204800KB) for virtualization
-IMAGE_ROOTFS_EXTRA_SPACE:append = "${@bb.utils.contains('DISTRO_FEATURES', 'virtualization', ' + 307200', '', d)}"
+# Add extra-space to rootfs (default: 716800KB) for virtualization in absence of flex-ewaol-baremetal
+IMAGE_ROOTFS_EXTRA_SPACE:append:feature-virtualization = "${@bb.utils.contains('DISTRO_FEATURES', 'flex-ewaol-baremetal', '', ' + 716800', d)}"
```
ahsanhussain, line 209: In absence of flex-ewaol-baremetal? Isn't it the opposite?

Secondly the "feature-virtualization" override won't be necessary if you're going to check on flex-ewaol-baremetal. 

P.S. have you confirmed that this works? flex-ewaol-baremetal doesn't land in the DISTRO_FEATURES until after the distro conf has been parsed, which this file is a part of.

ahsanhussain: Why exclude flex-ewaol-baremetal which is a superset of feature-virtualization? 

Secondly the exclusion will most likely not work because flex-ewaol-baremetal doesn't land in the DISTRO_FEATURES until after the distro conf is parsed, which this file is a part of.

hashraf91: ewaol feature adds extra space to RootFS. We checked that almost 1GB of free-space is available. So, we just need to add the space when virtualization feature is enabled but in absence of ewaol. Then we need space for tracing demo.

> Secondly the exclusion will most likely not work because flex-ewaol-baremetal doesn't land in the DISTRO_FEATURES until after the distro conf is parsed, which this file is a part of.

That is when you're including conf file conditioned on DISTRO_FEATURES but variable parsing seems to works. I checked the env.

ahsanhussain: For good measure should we make the check on USER_FEATURES variable instead, as the event for feature_overrides.bbclass is "ConfigParsed"

ahsanhussain (approved):

PR #196, opened 2024-12-30
Enable Flex OS builds on RedHat-8 and Rocky-8

hashraf91:
Replaced check on /etc/redhat-release with `lsb_release` as `lsb_release` is available on all DISTROs and is convenient for setup-environment script.

On Redhat/Rocky-8, at least GCC9 is required. The std::filesystems library in older GCC versions was located at non-standard paths and needed to be included manually for multiple recipes. The cleaner solution is to perform the build with GCC9 and hence gcc-toolset-9 is required.

JIRA-ID: SB-24005

### DIFF
--- a/meta-flex-distro/conf/distro/flex-os.conf
+++ b/meta-flex-distro/conf/distro/flex-os.conf
@@ -62,6 +62,8 @@ PACKAGE_CLASSES ?= "package_ipk"
 # Flex OS's supported hosts
 SANITY_TESTED_DISTROS = "\
     ubuntu-22.04 \n\
+    rhel*-8* \n \
+    rocky*-8* \n \
     rhel*-9* \n \
     redhatenterprise*-9* \n \
 "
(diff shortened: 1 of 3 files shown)


hashraf91:
I have tested the complete build on Redhat-8.2 and tested the packages installation and repositories on Rocky-8.9. Both setups were similar and use same repositories for newly installed packages. I haven't tested the build on Rocky-8.9.

Noor-Ahsan:
Apparently looks fine

ahsanhussain:
The commit message doesn't explain the reasoning for make43 and gcc-toolset-9

hashraf91:
> The commit message doesn't explain the reasoning for make43 and gcc-toolset-9

Okay, updated the commit message to explain why gcc-toolset-9 is installed. make43 was already addressed under JIRA: SB-23264

hashraf91:
This PR is also tested on Rocky-8.10.

ahsanhussain (approved):

PR #199, opened 2025-01-24
llvm_19.1.6: port latest LLVM version and fix the cross-compilation config LLVM_HOST_TRIPLE

hashraf91:
We are experiencing issue with EGL initialization. As mesa depends on LLVM for llvmpipe driver, so porting the latest version available as of today which seems to fix the issue.
Ported from:
https://github.com/openembedded/openembedded-core/commit/8546473f724e16095e8ad68813f3a087cb1b2779

Tested-by: Ahsan Hussain <ahsan.hussain@siemens.com>
Signed-off-by: Haseeb Ashraf <haseeb_ashraf@mentor.com>

The correct way to configure cmake for cross-compilation includes setting the LLVM_HOST_TRIPLE as well. This fixes the bug when LLVM is cross-compiled for AAarch64 but it was getting built with the triple of native build system instead of the TARGET_SYS ('aarch64-oe-linux' in my case).
```
No available targets are compatible with triple "x86_64-unknown-linux-gnu"
```
Documentation Ref:
https://github.com/llvm/llvm-project/blob/llvmorg-19.1.6/llvm/docs/HowToCrossCompileLLVM.rst#configuring-cmake

JIRA-ID: SB-24074

Signed-off-by: Haseeb Ashraf <haseeb_ashraf@mentor.com>

### DIFF
--- a/meta-flex-staging/recipes-devtools/llvm/llvm_%.bbappend
+++ b/meta-flex-staging/recipes-devtools/llvm/llvm_%.bbappend
@@ -1,0 +1,6 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+
+EXTRA_OECMAKE:append:class-target = " -DLLVM_HOST_TRIPLE=${TARGET_SYS}"
+EXTRA_OECMAKE:append:class-nativesdk = " -DLLVM_HOST_TRIPLE=${SDK_SYS}"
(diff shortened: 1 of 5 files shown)


meta-flex-staging/recipes-devtools/llvm/llvm_%.bbappend
```diff
@@ -0,0 +1,6 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------------------------------------------------
+
+EXTRA_OECMAKE:append:class-target = " -DLLVM_HOST_TRIPLE=${TARGET_SYS}"
+EXTRA_OECMAKE:append:class-nativesdk = " -DLLVM_HOST_TRIPLE=${TARGET_SYS}"
```
ahsanhussain: Why change for class-nativesdk? Isn't it for SDKMACHINE

hashraf91: class-nativesdk is for SDK which could also built for different arch. See the original recipe: https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-devtools/llvm/llvm_19.1.6.bb#L89

hashraf91: Okay, this should be SDK_SYS. Thanks for pointing this out.

ahsanhussain (approved):
LGTM, thanks!

ahsanhussain:
Merging as per https://teams.microsoft.com/l/message/19:fff74218dbde446088e4e88eb91cdb14@thread.v2/1737973659068?context=%7B%22contextType%22%3A%22chat%22%7D

PR #200, opened 2025-01-27
Graphics stack and vkmark upgrade

ahsanhussain:


meta-flex-distro/recipes-graphics/vkmark/vkmark/0001-src-meson.build-Prepend-sysroot-for-the-includedir.patch
```diff
@@ -8,22 +8,23 @@ Need to prepend PKG_CONFIG_SYSROOT_DIR to the dir path for cross compilation
...
  
 +pkgconf_sysroot = run_command(prog_python, '-c', 'import os; print(os.environ.get("PKG_CONFIG_SYSROOT_DIR"))').stdout().strip()
-+vulkan_hpp = pkgconf_sysroot + vulkan_hpp
-+
- format_map_gen_h = custom_target(
-     'format_map_gen.h',
-     output: 'format_map_gen.h',
+ vk_xml = join_paths([
+-    vulkan_dep.get_pkgconfig_variable('prefix'),
++    pkgconf_sysroot, 'usr',
```
hashraf91: better if we use this: `pkgconf_sysroot + vulkan_dep.get_pkgconfig_variable('prefix'),` as it will not override prefix from vulkan_dep and only prepend the pkgconf_sysroot.

ahsanhussain: That's not where the Vulkan registry goes, so it won't resolve to the correct path for the xml

meta-flex-staging/recipes-graphics/drm/libdrm_2.4.123.bb
```diff
@@ -0,0 +1,66 @@
+# This material contains trade secrets or otherwise confidential information owned by Siemens Industry Software Inc.
```
hashraf91: where should the backported recipes be placed? Previously you placed recipes in meta-flex-distro, so I also placed llvm recipe there. And placed only the fix in meta-flex-staging.

meta-flex-distro/recipes-graphics/vkmark/vkmark/0001-src-meson.build-Prepend-sysroot-for-the-includedir.patch
hashraf91: I have this tested ;). The join_paths functions such that if any of the argument has '/' in front i.e. an absolute path, it discards the previous arguments. If you were joining them with ',' then it was just ignoring the pkgconf_sysroot. It works if you use '+' for concatenation.

meta-flex-staging/recipes-graphics/drm/libdrm_2.4.123.bb
ahsanhussain: Ideally in meta-flex-staging, so we can remove from there when we update the Yocto version.

hashraf91: Okay, you're placing those recipes in meta-flex-staging which are anticipated to be available in future by Yocto update and remaining ones in distro which are just taken from other layers?

ahsanhussain: Kind of. So the recipes anticipated to be updated by the Yocto version update have been going in meta-flex-staging for a while now. That's why we removed a bunch from there when update to scarthgap

meta-flex-distro/conf/distro/innexis-linux.conf
```diff
@@ -329,5 +329,11 @@ include conf/distro/include/flex-distro.conf
 # Conditional EWAOL integration using USER_FEATURES because those wouldn't
 # have been added to DISTRO_FEATURES while parsing distro.conf
 require ${@bb.utils.contains('USER_FEATURES', 'flex-ewaol-baremetal', 'conf/distro/include/ewaol-flex-integration.conf', '', d)}
+
+# DISTRO preferences
+PREFERRED_VERSION_libdrm = "2.4.123"
+PREFERRED_VERSION_mesa   = "24.3.4"
+PREFERRED_VERSION_wayland-protocols = "1.38"
```
hashraf91, line 336: Do we need to force these version? As these will be picked up by default. Their native versions should also be forced to match such as for mesa-native.

meta-flex-distro/recipes-graphics/vkmark/vkmark/0001-src-meson.build-Prepend-sysroot-for-the-includedir.patch
ahsanhussain: OK, updated now. Didn't dig into it once the path wasn't resolving :)

meta-flex-distro/conf/distro/innexis-linux.conf
ahsanhussain: Unless selected in a high priority layers, these would stay at these versions. If we don't mention here then any recipe from a higher priority layer will override these versions, so it's a good means to have a working configuration in the conf for now. 

As for the native version, do we even care about that? native one doesn't build with any of the new packageconfigs, and also we don't use it so that's why I didn't touch it.

hashraf91 (approved):

ahsanhussain:
Pushed a commit for drm_info utility.

hashraf91 (approved):

meta-flex-staging/recipes-graphics/mesa/mesa.inc
```diff
@@ -0,0 +1,393 @@
...
+
+PACKAGECONFIG[broadcom] = ""
+PACKAGECONFIG[etnaviv] = ""
+PACKAGECONFIG[freedreno] = ""
+PACKAGECONFIG[kmsro] = ""
+PACKAGECONFIG[vc4] = ""
+PACKAGECONFIG[v3d] = ""
+PACKAGECONFIG[zink] = ""
+
+GALLIUMDRIVERS = "softpipe,llvmpipe"
```
hashraf91: This line should be split in to following:
```
GALLIUMDRIVERS = "softpipe"
GALLIUMDRIVERS_LLVM:append = ",llvmpipe"
```
because the llvmpipe now gets enabled by gallium-llvm packageconfig.
You can merge the second line with this: https://github.com/MentorEmbedded/meta-sokol-flex/pull/200/files#diff-1bee4e58bd8a7a0a2f2c496a602ca690819bc7435d2d2e43b18883f77e543962R173

hashraf91 (changes requested):

meta-flex-staging/recipes-graphics/mesa/mesa.inc
ahsanhussain: Keeping the default behavior of this recipe, this should stay here. gallium-llvm packageconfig only used to enable the Vulkan llvmpipe driver (lavapipe) and that's how it is now.

hashraf91: The behavior has changed. Previously, the recipe was directly depending upon llvm so enabling llvmpipe without any packageconfig was fine but now since the recipe depends on llvm only in case of gallium-llvm packageconfig, so the llvmpipe should only be enabled with that packageconfig.

ahsanhussain: Still depends explicitly on llvm. It was a missing llvm-native which was causing an issue for innexis hybrid BSP build. Thanks for testing this. I've added a commit.

hashraf91 (dismissed):

hashraf91 (approved):

PR #201, opened 2025-02-06
aemu: Move Mesa dependency from BSP layer

ahsanhussain:
Although currently only being used for ANA BSP, this is a Mesa 24.3 dependency which should live with the other Mesa changes that we did for the recipe upgrade. Move it to the distro layer


ahsanhussain:
Needs https://github.com/MentorEmbedded/meta-flex-ref-bsps/pull/285

PR #206, opened 2025-03-26
scripts: rename setup-rh -> setup-rocky

hashraf91:
We are decommissioning support for Redhat hosts for Innexis Linux builds. This script was previously used for both Redhat and Rocky hosts. Renaming the script name and adjusting the script to work for Rocky hosts only.

JIRA-ID: SB-24107


ahsanhussain:
The commit title is very informal. Let's add a commit message as well as to why this change is needed

hashraf91:
> The commit title is very informal. Let's add a commit message as well as to why this change is needed

Added commit message. But title seems okay to me. I have seen some examples in oe-core renaming files in similar way.

ahsanhussain (approved):
Well, technically it's not just a rename so a title like "setup-rh: remove support for RHEL hosts" would be better. Not a big deal though, now that we have the commit message.

PR #208, opened 2025-04-07
Support build on Ubuntu-24 host

hashraf91:
* innexis-linux: update list of supported hosts
We decommissioned support for Redhat hosts and added support for Ubuntu-24 host for Innexis Linux builds.
* scripts/setup-ubuntu: add fix/workaround for build on ubuntu-24 host
Ubuntu-24 enables apparmor by-default which restricts the permissions for bitbake binary. No proper solution is available as of today, see the discussion here:
https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2056555
Secondly, Ubuntu-24 uses UMASK 0027 as default now instead of 0022 but oe-core's sanity requires 0022:
https://github.com/openembedded/openembedded-core/blob/yocto-5.0.3/meta/classes-global/sanity.bbclass#L842
* scripts: update regex for matching ID and VERSION_ID in os-release
Ignore the double quote (") while matching ID or VERSION_ID, and check version until dot is matched.

JIRA-ID: SB-24317

Signed-off-by: Haseeb Ashraf <haseeb_ashraf@mentor.com>

### DIFF
--- a/scripts/setup-ubuntu
+++ b/scripts/setup-ubuntu
@@ -42,4 +42,21 @@ fi
 echo "Generating the en_US.UTF-8 locale"
 $SUDO locale-gen en_US.UTF-8
 
+# Workaround for ubuntu-24 issue [Allow bitbake to create user namespace](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2056555)
+if grep -Pq '^VERSION_ID=(?:|")24\.' /etc/os-release; then
+    echo "Adding Bitbake's AppArmor Profile to associate it with User Namespace profile"
+    echo "NOTE: This is a workaround for Ubuntu bug (https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2056555)"
+    echo "which gives unconfined access to bitbake binary. You may review the AppArmor Profile for Bitbake available"
+    echo "here: /etc/apparmor.d/bitbake"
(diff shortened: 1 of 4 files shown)


hijaz-siemens (approved):

scripts/setup-ubuntu
```diff
@@ -42,4 +42,18 @@ fi
 echo "Generating the en_US.UTF-8 locale"
 $SUDO locale-gen en_US.UTF-8
 
+# Workaround for ubuntu-24 issue [Allow bitbake to create user namespace](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2056555)
+if grep -Pq '^VERSION_ID=(?:|")24\.' /etc/os-release; then
```
ahsanhussain, line 46: This is an intrusive change to the host OS. Perhaps ask for user permission here, or we need to document it in the user facing docs.

hashraf91: Bitbake will not work without it, so I think we should not ask user when he actually doesn't have a choice to proceed without saying 'yes'. We can print out this information when script is run, so user is aware of it, thoughts?

hashraf91: Updated the message logs when setting up bitbake apparmor profile. Please check if this looks okay.

ahsanhussain (approved):

PR #210, opened 2025-04-30
mesa: fix vkmark related issues

hashraf91:
This fixes two issues.

1. mesa: fix vkmark unrealistic score for CPU-based rendering
    
    Port back the patches to replace aemu library provided allocator
    `android::base::SubAllocator` to `u_mm.h` allocation APIs.
    
    The vkmark reported score for host CPU-based lavapipe gfxstream ICD
    before this change was ~8000 and after this change it is ~1500.
    
    The latter score makes sense as the GPU-based gfxstream ICD gives
    a score of ~3500.
    
    Ref: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/32015
    
    JIRA-ID: SB-24465

2. mesa: fix vkmark `texture:anisotropy=16` benchmark intermittent hang
    
    When anisotropy is enabled i.e. sampling > 1, benchmark is observed
    to hang. The main patch that fixes this issue is:
    https://gitlab.freedesktop.org/mesa/mesa/-/commit/350a0fe63298d72593c473a0aae83afcd84de9b8
    
    Other patches fixes bugs introduced by this new implementation of
    anisotropy filter. These patches are ported back from mesa-25.
    
    Please note that these patches fixes the anisotropy issue with guest
    CPU-based llvmpipe usage only.
    
    JIRA-ID: SB-24083

### DIFF
--- a/meta-flex-staging/recipes-graphics/mesa/files/0001-llvmpipe-Disable-anisotropic-filtering-for-explicit-.patch
+++ b/meta-flex-staging/recipes-graphics/mesa/files/0001-llvmpipe-Disable-anisotropic-filtering-for-explicit-.patch
@@ -1,0 +1,28 @@
+From 1196559d894abcfa31e0c28281206aaa7ac496a1 Mon Sep 17 00:00:00 2001
+From: Konstantin Seurer <konstantin.seurer@gmail.com>
+Date: Fri, 27 Dec 2024 13:17:38 +0100
+Subject: [PATCH 1/9] llvmpipe: Disable anisotropic filtering for explicit lod
+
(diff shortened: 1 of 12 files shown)


ahsanhussain:
Which version of Mesa are these patches from? We can update the Mesa version instead of having to maintain this many patches.

hashraf91:
> Which version of Mesa are these patches from? We can update the Mesa version instead of having to maintain this many patches.

These are from mesa 25.0 branch mostly. I didn't update mesa for now because it would require testing of other features again. IMO, let's provide images to ANA team with this change. After this, we can upgrade to mesa-25 and drop these patches.

ahsanhussain:
Fine

PR #212, opened 2025-05-08
remove setup-debian script and Include standard dtb file for prebuild image

pragow0k:
Removing setup-debian script from installer
[SB-24473](https://jira.alm.mentorg.com/browse/SB-24473)

------------------------
Include KERNEL_DEVICETREE_ARCHIVE_RELEASE variable for dtb file in DEPLOY_IMAGES for archive release.
[SB-24472
](https://jira.alm.mentorg.com/browse/SB-24472)
KERNEL_DEVICETREE defined in, https://github.com/hijaz-siemens/meta-flex-ref-bsps/blob/7f75b5c97da26b7a4c70c2fe3bc2dbd921cbc97c/meta-flex-bsp-innexis-arm64/conf/machine/innexis-arm64.conf#L22

### DIFF
--- a/meta-flex-support/recipes-core/meta/archive-release.bb
+++ b/meta-flex-support/recipes-core/meta/archive-release.bb
@@ -100,9 +100,11 @@ SRC_URI:append:qemuall = " file://runqemu.in"
 # Image files to be archived
 IMAGE_BASENAME = "${RELEASE_IMAGE}"
 EXTRA_IMAGES_ARCHIVE_RELEASE ?= ""
+KERNEL_DEVICETREE_ARCHIVE_RELEASE ?= ""
 DEPLOY_IMAGES ?= "\
     ${@' '.join('${IMAGE_LINK_NAME}.%s' % ext for ext in d.getVar('IMAGE_EXTENSIONS').split())} \
     ${EXTRA_IMAGES_ARCHIVE_RELEASE} \
+    ${KERNEL_DEVICETREE_ARCHIVE_RELEASE} \
 "
 DEPLOY_IMAGES:append:qemuall = "${@' ' + d.getVar('KERNEL_IMAGETYPE') if 'wic' not in d.getVar('IMAGE_EXTENSIONS') else ''}"
 DEPLOY_IMAGES[doc] = "List of files from DEPLOY_DIR_IMAGE which will be archived"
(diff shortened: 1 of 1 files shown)


meta-flex-support/recipes-core/meta/archive-release.bb
```diff
@@ -100,9 +100,12 @@ SRC_URI:append:qemuall = " file://runqemu.in"
 # Image files to be archived
 IMAGE_BASENAME = "${RELEASE_IMAGE}"
 EXTRA_IMAGES_ARCHIVE_RELEASE ?= ""
+KERNEL_DEVICETREE ?= ""
```
hashraf91: we should not redefine this variable from a recipe scope

meta-flex-support/recipes-core/meta/archive-release.bb
```diff
@@ -100,9 +100,12 @@ SRC_URI:append:qemuall = " file://runqemu.in"
...
+DTB ?= "linux-std.dtb"
```
hashraf91: DTB is too generic variable name to use. Instead let's use `KERNEL_DEVICETREE_ARCHIVE_RELEASE`.

Preferably, we should simply include this variable in DEPLOY_IMAGES and then control it specific to BSP from machine.conf for dtb(s) we want to package for that BSP.

meta-flex-support/recipes-core/meta/archive-release.bb
pragow0k: changed: https://jira.alm.mentorg.com/browse/SB-24472#:~:text=changes%20after%20comment%3A

meta-flex-support/recipes-core/meta/archive-release.bb
pragow0k: changed: https://jira.alm.mentorg.com/browse/SB-24472#:~:text=changes%20after%20comment%3A

meta-flex-support/recipes-core/meta/archive-release.bb
```diff
@@ -103,6 +103,7 @@ EXTRA_IMAGES_ARCHIVE_RELEASE ?= ""
 DEPLOY_IMAGES ?= "\
     ${@' '.join('${IMAGE_LINK_NAME}.%s' % ext for ext in d.getVar('IMAGE_EXTENSIONS').split())} \
     ${EXTRA_IMAGES_ARCHIVE_RELEASE} \
+    ${@'linux-std.dtb' if 'innexis-arm64' in d.getVar('MACHINE') else ''} \
```
kergoth: This would be best done with an override, rather than inline python, but usually this sort of thing belongs in the machine configuration, or the machine config override file included by the distro config, rather than in the recipe itself.

pragow0k: updated

hashraf91 (approved):

PR #214, opened 2025-05-22
Dcook at1 3558 virtio sound enablement for innexis linux for ana

cookdave:
Implements review comment action from here:
https://github.com/MentorEmbedded/meta-flex-ref-bsps/pull/305/


cookdave:
didn't mean to try and merge with main! d'oh.

PR #217, opened 2025-06-02
staging: yocto-cfg-fragments: include devmapper configs in docker

hashraf91:
Adds following patch for yocto-cfg-fragments:

* cfg/docker.cfg enable pre-req of CONFIG_DM_THIN_PROVISIONING

    CONFIG_MD and CONFIG_BLK_DEV_DM are pre-requisite of
    CONFIG_DM_THIN_PROVISIONING.

JIRA-ID: SB-24462


ahsanhussain:
While you're at it can you also add CONFIG_EXT4_FS_SECURITY. This is not a requirement for runc/docker but some docker images expect this from the host FS driver, and fail to run containers without. This is enabled on other ewaol distros usually

ahsanhussain:
I had to rebuild flex kernel a bunch of times because of this

hijaz-siemens (approved):

hashraf91:
@ahsanhussain updated the PR, please check now.

ahsanhussain (approved):

PR #219, opened 2025-07-04
innexis-linux.conf: remove SDK customization for Innexis Codebench

hashraf91:
Innexis Codebench now supports upstream Yocto SDK. The Codebench metadata, extra variables and environment-setup.d hack bbclasses are not required.

JIRA-ID: SB-24575


meta-flex-distro/conf/distro/innexis-linux.conf
```diff
@@ -280,15 +280,8 @@ TOOLCHAIN_TARGET_TASK_EXTRA += " sdk-env-kernelvars"
...
 TOOLCHAIN_HOST_TASK:remove = "${TOOLCHAIN_HOST_REMOVE}"
 
-# Include metadata for CodeBench in the SDK
-SDKIMAGE_FEATURES:append = " codebench-metadata"
-IMAGE_CLASSES += "sdk_codebench_metadata"
-
 # We need to ensure we can distribute downloads for SDK builds
 IMAGE_CLASSES += "archive_sdk_downloads"
-
-# Create a 'latest' symlink for the SDK
```
Noor-Ahsan: I think this is useful thing for SDK. It does not add anything in the SDK but creates a symlink to the latest SDK that is being built. I think we should keep this.

meta-flex-common/classes/sdk_latest_link.bbclass
```diff
@@ -1,18 +0,0 @@
-# ---------------------------------------------------------------------------------------------------------------------
-# SPDX-License-Identifier: MIT
-# ---------------------------------------------------------------------------------------------------------------------
-
-TOOLCHAIN_LINKNAME ?= "${@'${TOOLCHAIN_OUTPUTNAME}'.replace('-${SDK_VERSION}', '')}"
```
Noor-Ahsan: we are setting Toolchain output name herehttps://github.com/MentorEmbedded/meta-sokol-flex/blame/scarthgap/meta-flex-distro/conf/distro/innexis-linux.conf#L268. Read the comment. It is looking odd to me. When we added back toolchain in SDK do we need to update it. Kindly check please

hashraf91: Yes, some more cleanup and adjustment is required in innexis-linux.conf. I put this PR in draft state.

Noor-Ahsan: if change is fine then atleast we should update the comment

hijaz-siemens (approved):

meta-flex-common/classes/sdk_latest_link.bbclass
hashraf91: Updated the comment.

meta-flex-distro/conf/distro/innexis-linux.conf
hashraf91: Okay, brought back this change.

PR #224, opened 2025-10-29
innexis-linux.conf: changing builddir to topdir

hijaz-siemens:
The BUILDDIR variable is not available to recipes in Yocto 5.0.11 so changing it to TOPDIR for flex external artifacts.

JIRA-ID: SB-24780


hashraf91 (approved):

Noor-Ahsan:
just for reference reviously we were using topdir but we moved to builddir in this commit https://github.com/MentorEmbedded/meta-sokol-flex/commit/924482522740cbc722b60a013fe6d4299fd87731

PR #227, opened 2026-01-19
setup-rocky: remove tar from rocky setup script.

pragow0k:
info: tar package was upgraded when setup script
        was ran which was causing build failure.

### DIFF
--- a/scripts/setup-rocky
+++ b/scripts/setup-rocky
@@ -11,7 +11,7 @@
 
 packages="python39 \
           python3-setuptools \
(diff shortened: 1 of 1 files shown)


Noor-Ahsan:
can you update the commit message. Summary should start from "setup-rocky:" instead of "[SB-24904]:"

pragow0k:
> can you update the commit message. Summary should start from "setup-rocky:" instead of "[SB-24904]:"

updated

PR #230, opened 2026-03-18
weston-init.sh: export mesa_vk_wsi_debug

hijaz-siemens:
vkmark reports an error on jasper-lite:
*ERROR* response 0x1200
Need to export MESA_VK_WSI_DEBUG=sw,buffer

JIRA-ID: SB-24949


hashraf91 (approved):

Noor-Ahsan:
you are fixing it on sokol flex. this will be applicable on devpro too. are you sure that we need to fix it on devpro as well

PR #232, opened 2026-07-03
Integrate support for Virtual Machine Managers (Xen and KVM) for Dolphin-9 VSOC

hashraf91:
* mesa: synchronize with upstream recipe and upgrade to 25.3.6
* distro: integrate meta-telechips-support layer
* gst-example: fix backporting errors
* distro: allow enabling of distro feature gobject-introspection-data
* glmark2: add packageconfig gbm-gl and gbm-gles2
* libglvnd: backport recipe version 1.7.0 and enable glvnd
* virglrenderer: add version 1.2.0
* pixman: fix arm64 advanced prefetcher bug
* aemu: update version to match ANA virtio-gpu-gfxstream backend
* gfxstream: add gfxstream recipe
* rust: add versions 1.81.0
* rutabaga_gfx_ffi: add as yocto recipe
* libmks: add version 0.1.5
* tmux-xpanes: add recipe
* qemu: add version 10.2.2
* qemu: enable pipewire backend for virtio-sound
* qemu: virgl: support async fencing feature of virglrenderer
* xen,xen-tools: add version 4.21
* xen: add fixes to avoid stalls during DomU re-creation
* Add distro feature for Virtual Machine Managers


meta-flex-distro/conf/distro/innexis-linux.conf
```diff
@@ -313,6 +358,9 @@ FLEX_EXTERNAL_ARTIFACTS ?= "${TOPDIR}/flex-external-artifacts"
 # ML configuration
 include conf/distro/include/flex-ml.conf
 
+# Telechips configuration
+include conf/distro/include/telechips-support.conf
+
```
ahsanhussain, line 363: Instead of adding customer specific include in distro conf, can't you use flex-distro.conf from telechips layer at higher priority so it gets picked

hashraf91: Then flex-distro.conf from BSP layer won't be picked up.

